PR #9, opened 2016-11-18
Add method to create a background request for /asset/v3 uploads

daehn:
# What's in this PR?

* The upload requests using the `/assets/v3` endpoint were not using the background `URLSession` and thus immediately terminated when the app was killed or the network connection lost.

### DIFF
--- a/WireRequestStrategy/Helper/AssetRequestFactory.swift
+++ b/WireRequestStrategy/Helper/AssetRequestFactory.swift
@@ -33,7 +33,8 @@ public final class AssetRequestFactory : NSObject {
     public func backgroundUpstreamRequestForAsset(message: ZMAssetClientMessage, withData data: Data, shareable: Bool = true, retention: Retention = .persistent) -> ZMTransportRequest? {
         let path = "/assets/v3"
         guard let uploadURL = uploadURL(for: message, in: message.managedObjectContext!, shareable: shareable, retention: retention, data: data) else { return nil }
-        let request = ZMTransportRequest.uploadRequest(withFileURL: uploadURL, path: path, contentType: "multipart/mixed")
+        let request = ZMTransportRequest.uploadRequest(withFileURL: uploadURL, path: path, contentType: "multipart/mixed; boundary=frontier")
+        request.addContentDebugInformation("Uploading full asset to /assets/v3")
         return request
     }
 
--- a/WireRequestStrategy/Helper/AssetRequestFactory.swift
+++ b/WireRequestStrategy/Helper/AssetRequestFactory.swift
@@ -25,18 +25,25 @@ public final class AssetRequestFactory : NSObject {
     let octetStreamContentType = "application/octet-stream"
     
     public enum Retention : String {
-        case Persistent = "persistent"
-        case Eternal = "eternal"
-        case Volatile = "volatile"
+        case persistent = "persistent"
+        case eternal = "eternal"
+        case volatile = "volatile"
     }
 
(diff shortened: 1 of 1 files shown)


WireRequestStrategy/Helper/AssetRequestFactory.swift
```diff
@@ -46,5 +53,10 @@ public final class AssetRequestFactory : NSObject {
             ZMMultipartBodyItem(data: data, contentType: octetStreamContentType, headers: fileDataHeader),
             ], boundary: "frontier")
     }
+
+    private func uploadURL(for message: ZMAssetClientMessage, in moc: NSManagedObjectContext, shareable: Bool, retention: Retention, data: Data) -> URL? {
+        guard let multipartData = try? dataForMultipartAssetUploadRequest(data, shareable: shareable, retention: retention) else { return nil }
+        return moc.zm_fileAssetCache.storeRequestData(message.nonce, data: multipartData)
```
mikeger: When the stored data would be cleared? I guess this request, as it runs on background, can be completed by OS itself, giving us the callback on the next app start, so we should also there delete the unnecessary data.

WireRequestStrategy/Helper/AssetRequestFactory.swift
```diff
@@ -25,18 +25,25 @@ public final class AssetRequestFactory : NSObject {
...
+        case persistent = "persistent"
+        case eternal = "eternal"
+        case volatile = "volatile"
     }
 
-    public func upstreamRequestForAsset(withData data: Data, shareable: Bool = true, retention : Retention = .Persistent) -> ZMTransportRequest? {
+    public func backgroundUpstreamRequestForAsset(message: ZMAssetClientMessage, withData data: Data, shareable: Bool = true, retention: Retention = .persistent) -> ZMTransportRequest? {
+        let path = "/assets/v3"
+        guard let uploadURL = uploadURL(for: message, in: message.managedObjectContext!, shareable: shareable, retention: retention, data: data) else { return nil }
+        let request = ZMTransportRequest.uploadRequest(withFileURL: uploadURL, path: path, contentType: "multipart/mixed")
```
mikeger: I'm not an expert on this part of the protocol, but noticed that `boundary=frontier` is not included in here. Is is needed? Or `ZMTransportRequest.uploadRequest` appends it?

WireRequestStrategy/Helper/AssetRequestFactory.swift
daehn: The data is cleared in `AssetV3FileUploadRequestStrategy` once the request completes

mikeger: Cannot find it (AssetV3FileUploadRequestStrategy), should be in this repo?

WireRequestStrategy/Helper/AssetRequestFactory.swift
daehn: I was also drawn off by this as we do not put in the boundary for our v2 file uploads as well, `ZMTransport` should not modify the `content-type`, I will check this as this should not work for v2 as well as the boundary is needed for the receiver to know where the new item in the multipart starts.

WireRequestStrategy/Helper/AssetRequestFactory.swift
daehn: No that's in `wire-ioo-message-strategy`: https://github.com/wireapp/wire-ios-message-strategy/blob/develop/Source/Transcoders/AssetV3FileUploadRequestStrategy.swift#L255

daehn: Fixed now, but I am still couldn't find the place were we assume the boundary is `"frontier"`, as we do not specify it for v2 assets...

PR #18, opened 2017-03-13
Update for changed delegate method name

vytis:


daehn:
As a dependency was bumped in the major version, this should also be a major release.

vytis:
Will do after the merge

PR #21, opened 2017-03-20
Fix/refactor dependent objects

marcoconti83:
# Reason for this pull request
Dependency between objects to uploaded were tracker in two different structure, which had some code duplication: `DependencyMap` and `DependentObjects`. 
My guess is that `DependencyMap` was added to have Swift type safety, while `DependentObjects` is Obj-c compatible (and thus not type-safe). @typfel probably knows more on why `DependencyMap` was introduced.

# Changes
- `DependentObjects` has been made generic so that it now provides the type safety that `DependencyMap` was providing. To make it Objc-c compatible, I created a thin wrapper that just strips the type information.
- `DependentObjects` was using a NSMapTable, but in Swift we can safely use dictionaries. I also added a reverse mapping to speed up look-up by value.


typfel:
`DependencyMap` was introduced for the type safe `DependencySync` with the aim of using it as base for `UpstreamSync`and `UpstreamModifiedSync` which currently contains code duplication.

WireRequestStrategy/DependentObjects.swift
```diff
@@ -0,0 +1,119 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+fileprivate let zmLog = ZMSLog(tag: "Dependencies")
+
+
+public class DependentObjects<OBJECT: Hashable, DEPENDENCY: Hashable> {
```
daehn: I would change the casing from upper to capitalized for `Object` and `Dependency` like the standard library does for generics as well.

typfel:
@marcoconti83 Is this PR fixing a bug or ist just re-factoring?

marcoconti83:
@typfel the original intent was to add logs for debugging a bug with dependencies on internal builds. Once I started adding log to places where we were adding a dependency, I discovered there were more than one place were we were doing so, so I refactored to remove code duplication.

WireRequestStrategy/DependentObjects.swift
```diff
@@ -0,0 +1,119 @@
...
+public class DependentObjects<Object: Hashable, Dependency: Hashable> {
+    
+    public init() {
+    }
+    
+    private var dependenciesToDependents: [Dependency: Set<Object>] = [:]
+    private var dependentsToDependencies: [Object: Set<Dependency>] = [:] // inverse of the previous one
+ 
+    /// Adds a Dependency to an
+    public func add(dependent: Object, dependency: Dependency) {
```
typfel: I think this should have the same naming pattern as the `remove` method.

like:
`remove(dependency:for:)`
`add(dependency:for:)`

marcoconti83: Fixed

WireRequestStrategy/DependencyEntitySync.swift
```diff
@@ -66,29 +37,29 @@ public class DependencyEntitySync<Transcoder : EntityTranscoder> : NSObject, ZMC
         self.context = context
     }
     
-    public func expireEntities(withDependency dependency: AnyObject) {
-        for entity in entitiesWithDependencies.entities(withDependency: dependency) {
+    public func expireEntities(withDependency dependency: NSManagedObject) {
```
typfel: isn't it enough for `dependency`to be `Hashable`?

marcoconti83: True

PR #22, opened 2017-04-04
Fix/xcode8.3

marcoconti83:


daehn:
@marcoconti83 did you update the config.txt as well? COuldn't find it in the diffs.

vytis:
Update the Xcode version in circle.yml as well

PR #24, opened 2017-04-24
Some API changes required by state machine re-factoring

typfel:


daehn:
@typfel which dependencies does this PR have?

typfel:
@daehn It doesn't have any dependencies

PR #25, opened 2017-04-25
Add ptr to debug output for dependency

marcoconti83:
# Reason for this pull request
Currently in the debug logs for dependencies, it's not clear which dependency is cleaned. 
This PR adds the pointer to the object (it's easier said than done, in Swift) in those logs.


mikeger:
@vytis should we remove the Travis from the build? Looks like CircleCI is nailing it.

PR #27, opened 2017-05-17
Add helper extension to [ZMRequestGenerator]

daehn:
# What's in this PR?

* Makes it easier to call `nextRequest` on an array of potential generators.


vytis:
Maybe remove travis and move danger to Circle while you're at it?

WireRequestStrategy/Array+RequestGenerator.swift
```diff
@@ -0,0 +1,26 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
```
vytis: Works!

PR #29, opened 2017-05-24
Forward the response to the downstream transcoder

daehn:
# What's in this PR?

* In some cases we might not want to delete the object when `deleteObject:downStreamSync` is called on a `ZMDownstreamTranscoder`.
* We now pass the `ZMTransportResponse` down to the transcoder so that it can perform additional checks on the status code for example.


daehn:
Not sure if this should go to develop though, but if it does it should be a major release. As we don't change this framework too much I think it should be fine to merge it into develop and to do a major release.

PR #41, opened 2018-02-06
Cleaned up project structure

vytis:
### Issues

The project structure got cluttered quite a bit during the years and needed some clean up

### Solutions

With @typfel we tried to move files in folders in a way it makes more sense. This framework basically includes various types of syncs and their helpers, so the structure reflects this.

Code was moved into `Sources/` directory and relevant test files also moved together with the classes they are testing.

It turned out like this:

```
Helpers/
Protocols/
Image Preprocessing/
Other Syncs/
Request Syncs/
Object Syncs/
.../Helpers/
.../Protocols/
.../Upstream/
....../Helpers/
.../Downstream/
....../Helpers/
```

### Attachments

The current structure in Xcode now looks like this:

screen shot 2018-02-06 at 18 13 52
screen shot 2018-02-06 at 18 14 30


mikeger:
Should we establish a linting in danger to check if the structure is correct?

vytis:
I think the structure will be quite different for all frameworks, not sure if we can devise a general scheme that will work for everything...

PR #44, opened 2018-04-03
Merge framework with WireMessageStrategy

vytis:
### Issues

To have less confusion where is what we have decided to merge two frameworks into one.
It is possible to preserve all commit history so it's not a big problem - we will just need to change imports in various other places (mainly Sync Engine).

### Note

Make sure not to squash this when merging!


Resources/Configurations/WireMessageStrategy-ios.xcconfig
```diff
@@ -0,0 +1,2 @@
+#include "WireMessageStrategy.xcconfig"
+#include "zmc-config/ios.xcconfig"
```
typfel: Guess we only need one set of xcconfig files.

vytis: Oh yeah, need to delete the others

vytis:
@typfel updated by amending last commit

PR #45, opened 2018-04-04
Update asset retention policies

daehn:
## What's new in this PR?

* Update asset retention policies.

### This PR depends on:

- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/467 has been merged.


Sources/Helpers/AssetRequestFactory.swift
```diff
@@ -16,43 +16,62 @@
...
+        /// The asset will never be removed from the backend storage unless the
+        /// user requests the deletion explicitly. Used for profile pictures.
+        case eternal = "eternal"
+        
+        /// The same as eternal, however this is cost-optimized
+        /// on the backend for infrequent access. Used for team conversations.
+        case eternalInfrequentAccess = "eternal-infrequent_access"
+    }
+    
+    private struct Constant {
```
vytis: I think these sort of things are better to be enums, in that case you can't do `let a = Constant()` and things like that

daehn: @vytis very good point, updated!

PR #46, opened 2018-04-04
Fix/markdown link preview

johnxnguyen:
### Issues

We don't want to generate a link preview for markdown links.

### Causes

Even though the URL of a markdown link is hidden from the user, it is still detected when processing the message and a link preview is generated (if this link is the first link in the message).

### Solutions

`LinkPreviewDetector` (see [this](https://github.com/wireapp/wire-ios-link-preview)) populates an array of detected links within a string. Currently, a link preview is generated for only the first element. Using the new `LinkPreviewDetectorDelegate`, we are now able to decide which detected links will be included in this array, and we exclude those that are detected within the markdown syntax `[link name](url)`.

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-link-preview/pull/24

### DIFF
--- a/Sources/Request Strategies/Assets/Helpers/LinkPreviewPreprocessor.swift
+++ b/Sources/Request Strategies/Assets/Helpers/LinkPreviewPreprocessor.swift
@@ -36,6 +36,7 @@ private let zmLog = ZMSLog(tag: "link previews")
         self.linkPreviewDetector = linkPreviewDetector
         self.managedObjectContext = managedObjectContext
         super.init()
+        (self.linkPreviewDetector as? LinkPreviewDetector)?.delegate = self
     }
 
     public func objectsDidChange(_ objects: Set<NSManagedObject>) {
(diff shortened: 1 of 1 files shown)


Sources/Request Strategies/Assets/Helpers/LinkPreviewPreprocessor.swift
```diff
@@ -36,6 +36,7 @@ private let zmLog = ZMSLog(tag: "link previews")
         self.linkPreviewDetector = linkPreviewDetector
         self.managedObjectContext = managedObjectContext
         super.init()
+        (self.linkPreviewDetector as? LinkPreviewDetector)?.delegate = self
```
mikeger: Is anything keeping us from adding the reference to the delegate to the protocol `LinkPreviewDetectorType` ?

johnxnguyen: nope, but I like this idea of passing it in as an argument to the method!

PR #47, opened 2018-04-10
ZIOS-8814: Check if message ID is valid when creating request

alexaubry:
### Issues

When the client receives a message with an invalid message identifier (not an UUID), the app will try to insert it at every launch and perpetually crash.

### Causes

This was caused by forced optional unwrapping in the message transcoder. When the message ID is invalid, the request factory returns nil. Unwrapping it, in this case, caused a crash.

### Solutions

We check if the request is a valid object before using it. If the request cannot be created, we return `nil` so that the strategy can discard the message.


vytis:
Maybe it would be safer to handle this a bit lower down the stack in `wire-transport`? If we receive an invalid message like this it's better to drop it immediately, because there might be quite a few other places where we access conversation id and probably crash in the same way.

PR #58, opened 2018-06-21
 Add safe guards against sending out wrong message

typfel:
### Issues

It's been observed in rare circumstances that the iOS client re-sends a message from another user.

### Causes

A message is only sent out if the `delivered` property is set to false, which should never be case the for received messages. However in theory a message could be received, the context saves and the app crashes before the `delivered` property has been updated to true.

### Solutions

-  Filter messages which do not have the sender equal to the self user.
- Mark a message immediately after creating it to prevent any saves happening in between.


typfel:
_All SE test passed with these changes._

PR #61, opened 2018-07-11
Re-add ImageDownloadRequestStrategyTests

daehn:
## What's in this PR?

* Re-add `ImageDownloadRequestStrategyTests` which were deleted previously.
* Only download version < 3 asset messages in case `message.hasEncryptedAsset` evaluates to true for them.


daehn:
@vytis checking locally now.

daehn:
Something weird though, test run fine locally for me. Re-running on CI.

mikeger:
@daehn was it removed for the reason? :)

daehn:
@mikeger yeah maybe, but I think it would be good to have these tests around.

PR #62, opened 2018-07-17
Hack to fix flaky test on SyncEngine

vytis:
### Issues

Several tests were constantly failing (e.g. https://circleci.com/gh/wireapp/wire-ios-sync-engine/3701). It was possible to reproduce the failure locally, but then it would fix itself after a while. After thorough cleanup (derived data, simulator and Carthage) it would come back again.

### Causes

While debugging the tests with WireRequestStrategy code pulled in we could see that it was mistakenly not deleting an outdated UserClient after getting the response. The offending line was:

```
<at this point: newClients.count == 0> 
newClients.formUnion(clients.filter { !$0.hasSessionWithSelfClient })
<at this point: newClients.count == 1> 
```

However in the debugger:

```
po clients.filter { !$0.hasSessionWithSelfClient }
0 items
```

On top of that if we try to step inside the `UserClient.hasSessionWithSelfClient` method it would instead step into `var activationLocation: CLLocation`. Something mysterious is clearly happening, but the reason is unclear. 

### Solutions

Rewriting the code to an equivalent one without using the `filter` seems to be fixing the issue. This is a weird hack, but after digging into the issue for two days this is the best bet...


vytis:
@zenkins release patch

PR #63, opened 2018-07-18
Fix flaky tests

vytis:
### Issues

Recently tests started failing on CI, but working fine locally. 

### Causes

We do a lot of switching between sync and ui MOC and this can cause unreliable tests.

### Solutions

Turned on CoreData concurrency debug flag which panics if we access things on wrong thread. Fixed all tests

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/520


vytis:
@zenkins release patch

PR #66, opened 2018-07-30
Fix a crash when creating link preview strategy

alexaubry:
### Issues

We force-unwrapped the current operation queue from the factory method that creates the link preview strategy, which caused a crash when it was called outside of the context of an `NSOperation`.

### Solutions

If the current operation queue is not available, we use the main operation queue as a fallback.


alexaubry:
@zenkins release patch

PR #68, opened 2018-08-10
Treat images larger than 5 MB as files

daehn:
## What's new in this PR?

* Treat images larger than 5 MB as files.

### This PR depends on:

- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/535 has been merged.


typfel:
There's currently no tests for storing a raster image `AssetV3DownloadRequestStrategy`. It is however almost identical to the file case, in fact I think we can remove the separation. It will require quite some re-factoring though so I wouldn't do it this PR.

typfel:
@zenkins release patch

PR #70, opened 2018-08-28
(RS) ZIOS-10377 - Assets sending is being constantly retried if sent in bad network conditions	

nicorsm:
### Issues

The text message sending expires in 20-30 seconds and Retry label appears underneath, but the image is still in Sending state and never gets the Retry option

### Causes

This is because images don't get expired after the deadline of 30 seconds.

### Solutions

I've changed `MessageExpirationTimer` class in order to allow expiration timers from `ZMClientMessage` and `ZMAssetClientMessage` classes.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/547

## Notes

This is the first implementation. Right now the expiration timer doesn't get updated on upload progress updates, causing the asset to be expired and then correctly uploaded. Any idea on how solving this issue?


daehn:
> Right now the expiration timer doesn't get updated on upload progress updates, causing the asset to be expired and then correctly uploaded.

Can you elaborate a bit more? We continue uploading even after we expired the message? I assume we should cancel the uploading process when the timer fires but I might be misunderstanding the issue.

nicorsm:
@daehn With this temporary implementation, yes. If I call `expire()` the asset continues to be uploaded when the network is back. From my point of view, it could easily be a concurrency thing that I don't consider, since I don't have the complete picture of how this works.

My original idea was to expire the upload if there are no new progress updates for the default timeout interval (30s). For this purpose, I should track the progress and postpone the expire date.

Sources/Request Strategies/Assets/AssetV3ImageUploadRequestStrategyTests.swift
```diff
@@ -78,6 +78,19 @@ class AssetV3ImageUploadRequestStrategyTests: MessagingTestBase {
         return message
     }
     
+    func createAssetImageMessage() -> ZMAssetClientMessage {
+        var message: ZMAssetClientMessage!
+        syncMOC.performGroupedBlockAndWait {
+            //self.conversation.messageDestructionTimeout = .local(.tenSeconds)
```
vytis: Commented out code, not sure if it should stay or be deleted

nicorsm:
@zenkins release major

PR #72, opened 2018-08-29
Update participant handling

typfel:
## What's new in this PR?

Use new `addParticipantIfMissing` added in https://github.com/wireapp/wire-ios-data-model/pull/549

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/549


typfel:
@zenkins release minor

PR #74, opened 2018-08-31
Feature/unordered relations masked

mikeger:
### Issues

Fixed tests for unordered messages.


Sources/Helpers/ZMMessage+Dependency.swift
```diff
@@ -90,7 +90,7 @@ extension ZMMessage {
         // so we iterate backwards and we ignore everything until we find this one
         var selfMessageFound = false
 
-        conversation.messages
+        NSOrderedSet(array: conversation.recentMessages)
```
vytis: Doesn't `enumerateObjects` work on array as well? Maybe we don't need to create a set here?

typfel: Bump

typfel: Couldn't you do normal swift enumeration and call `break` when you find your message.

mikeger: Great idea, updated it according to it!

typfel:
@zenkins release major

PR #75, opened 2018-08-31
save context after modifiying keys to trigger fetch

johnxnguyen:
## What's new in this PR?

We should save the context after setting the modified keys to trigger the fetch for the missing clients immediately.

Posting a `RequestsAvailableNotification` instead ought to do the trick, but strangely (and inexplicably) the observers didn't always receive the notification, even though the observers do not listen for notifications from a specific object, and the notifications aren't posted with a sender object. 

Even more strangely, the reason why saving the context works is because `ZMOperationLoop` in sync engine responds to the context save by posting its own `RequestsAvailableNotification`, which, as an observer of this notification, receives.

If anyone has an idea why this is, I'd love to hear it 😄


typfel:
I actually can't reproduce this issue on latest dev but It doesn't really hurt to save to if we change it into a `enqueueDelayedSave`.

mikeger:
@zenkins release patch

PR #76, opened 2018-09-06
Fix save loop

typfel:
### Issues

SE tests gets stuck waiting for all groups to finish when a client is missing.

### Causes

Doing a save will cause the operation loop to execute objectsDidChange on all context trackers. If a client is missing we will end up in `OTREntity.dependentObjectNeedingUpdateBeforeProcessingOTREntity` which always attempt to do:
```
 selfClient.setLocallyModifiedKeys(Set(arrayLiteral: ZMUserClientMissingKey))
 context.enqueueDelayedSave()
```

This will cause the operation loop to never to stop until the missing client is fetched.

### Solutions

Only update modified keys and save if the key is actually missing from the set.


typfel:
@zenkins release patch

PR #79, opened 2018-09-18
Unarchive the conversation when mentioned

mikeger:
### Issues

When current user is mentioned, we have to unarchive the conversation in order to bring it to the user's attention.


Sources/Request Strategies/Client Message/ClientMessageTranscoder.swift
```diff
@@ -151,6 +151,16 @@ extension ClientMessageTranscoder {
                 }
             }
             
+            // Unarchive the conversation, if current user is mentioned.
+            if let updateMessage = updateResult.message,
```
typfel: This does not account for the cleared timestamp. Existing logic for unarchiving on an incoming message is done in ZMOTRMessage.m:257

Maybe that code could be cleaned up and ported to Swift.

mikeger: Makes sense, will port it to data model.

PR #82, opened 2018-09-28
Process generic message for notifications first

johnxnguyen:
### Issues

When a message is deleted by the sender, the local notification on the receiver side is not removed from the lock screen / Notification Center. 

### Causes

Notification cancellation for deleted or edited messages is done by the `LocalNotificationDispatcher` in `process(genericMessage:)` (see [SyncEngine](https://github.com/wireapp/wire-ios-sync-engine/blob/f63ab47393404b044840ba37729f18379ed7dd60/Source/Notifications/Push%20notifications/LocalNotificationDispatcher%2BMessages.swift#L46)). The problem is, that update result created in this situation is nil (see [DataModel](https://github.com/wireapp/wire-ios-data-model/blob/c3da37dfa4d2053f7cac152a3808faf8da8d9cdb/Source/Model/Message/ZMOTRMessage.m#L173)). As a result, we return early, and the local notification dispatcher is not able to process the generic message.

### Solutions

Process the generic message first, then process with the update result.


mikeger:
@zenkins release patch

PR #85, opened 2018-10-29
Added completion handler for delivery receipts

vytis:
### Issues

When processing push notifications we were calling the completion handler earlier than needed - we were still sending delivery receipts. This might lead to app termination because we still are processing messages, but we tell to iOS that we are done.

### Solutions

Added a possibility to register completion handler which is called when all delivery receipts have been already sent.


vytis:
Might not be needed in final implementation

PR #88, opened 2018-11-28
Strategy for syncing read receipts

mikeger:
### Issues

Sync strategy to set the value upstream and respond to the downstream event. What is missing is 

- [x] Response failure handling.
- [x] Tests.
- [ ] Slow sync. -> Next PR


Sources/Request Strategies/Properties/UserPropertyRequestStrategy.swift
```diff
@@ -0,0 +1,181 @@
...
+        return nil
+    }
+    
+    public func updateUpdatedObject(_ managedObject: ZMManagedObject, requestUserInfo: [AnyHashable : Any]? = nil, response: ZMTransportResponse, keysToParse: Set<String>) -> Bool {
+        return false
+    }
+    
+    public func shouldRetryToSyncAfterFailed(toUpdate managedObject: ZMManagedObject, request upstreamRequest: ZMUpstreamRequest, response: ZMTransportResponse, keysToParse keys: Set<String>) -> Bool {
+        // Todo: reset value
+        return false
```
mikeger: @typfel I guess we cannot really set the property to the previous value, so we have to retry the request here? Could be unsafe, i.e. lead to request loops. Is this generated only on permanent error?

mikeger: Yeah, it's called when state is `ZMTransportResponseStatusPermanentError`.

typfel: Ok, so the only option is to give up I guess. This should in theory never happens I guess.

mikeger: Yeah.

mikeger:
@typfel updated

PR #90, opened 2018-11-29
Read receipts

typfel:
## What's new in this PR?

- Support changing read receipt account setting
- Support for requesting read receipt when sending message in one-to-one conversations


typfel:
@zenkins release major

PR #92, opened 2018-12-03
Use both ObjC and Swift protobuf bindings

vytis:
## What's new in this PR?

This PR fixes issues when building with Swift protobuf support

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-protos/pull/27
- [x] https://github.com/wireapp/wire-ios-data-model/pull/618


vytis:
@alexaubry maybe that would make more sense actually, just did not want to change wire-ios-link-preview for something that is not really a dependency.

vytis:
@zenkins release major

vytis:
@zenkins release major

PR #93, opened 2018-12-03
User properties slow sync

mikeger:
### Issues

Slow sync of the user properties: in case the app was offline for a while and did not managed to fetch the user settings, or just updated to the new version and have to fetch the current values.

Also updated the WIRE_ENABLE_READ_RECEIPTS format for the new spec.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/622


Sources/Request Strategies/Properties/UserPropertyRequestStrategy.swift
```diff
@@ -55,35 +55,99 @@ extension UserProperty {
...
+    }
+    
+    func parseUpdate(for selfUser: ZMUser, updateType: UpdateType, payload value: Any?) {
+        switch (self, updateType) {
+        case (.readReceiptsEnabled, .set):
+            let intValue: Int
+            if let numberValue = value as? Int {
+                intValue = numberValue
+            }
+            else if let stringValue = value as? String,
```
typfel: When would this happen?

mikeger: I don't know how the JSON decoder would treat the BE response, also this is the case in the test 🦀

mikeger: Now testing against the real BE

mikeger:
@typfel please approve this one

mikeger:
@zenkins release minor

Sources/Request Strategies/Properties/UserPropertyRequestStrategy.swift
typfel: So you'll update this later when you know for certain what the response type is?

mikeger: I think it's fine like that, since it's pretty robust against the wrong values, but I can also update it to make sure it's working one way. Had no chance to test with real BE, since the UI is always incompatible with frameworks because of other major changes.

PR #94, opened 2018-12-04
Fix: request loop on property 404

mikeger:
### Issues

The permanent error case was not really handled.


mikeger:
@zenkins release patch

PR #97, opened 2018-12-07
Set readReceiptsEnabledChangedRemotely

mikeger:
### Issues

This is necessary in order to show user that the setting was changed remotely. Flag is displaying the dot on the user info and generating the alert.

Depends on:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/627

### DIFF
--- a/Sources/Request Strategies/Properties/UserPropertyRequestStrategy.swift
+++ b/Sources/Request Strategies/Properties/UserPropertyRequestStrategy.swift
@@ -65,7 +65,14 @@ extension UserProperty {
         return ZMTransportRequest(getFromPath: path)
     }
     
-    enum UpdateType {
+    typealias  UpdateType = (source: UpdateSource, method: UpdateMethod)
+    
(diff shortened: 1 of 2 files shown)


Sources/Request Strategies/Properties/UserPropertyRequestStrategy.swift
```diff
@@ -65,7 +65,14 @@ extension UserProperty {
         return ZMTransportRequest(getFromPath: path)
     }
     
-    enum UpdateType {
+    typealias  UpdateType = (source: UpdateSource, method: UpdateMethod)
+    
+    enum UpdateSource {
```
typfel: Not suggesting you should change in this PR but this information is also available on the `ZMUpdateEvent`. We also usually create an `ZMUpdateEvent` from the HTTP response so that we can unify the processing.

mikeger:
@typfel please (finish the) review

mikeger:
@zenkins release major

PR #98, opened 2018-12-13
Fix read images expire after being sent

typfel:
### Issues

Image expire if they are read

### Causes

Another case of not taking the new .read delivery state into account.

### Solutions

Also check for the read state.

## Notes

As already mentioned in https://github.com/wireapp/wire-ios-data-model/pull/637 we should avoid these kind of bugs in the future. I think a good way would be to expose something like a `isDelivered` boolean which already exists on `ZMOTRMessage`. I'll not write a test for this now since I think we should re-factor this right away.


typfel:
@zenkins release patch

PR #103, opened 2019-01-18
Update after timers behaviour change

vytis:
### Issues

Ephemeral timers are not created explicitly after changes https://github.com/wireapp/wire-ios-data-model/pull/648 and some tests fail because of that.

### Solutions

Explicitly create timers in base test class.


vytis:
@zenkins release major

PR #104, opened 2019-01-21
Fix after transport changes

vytis:
## What's new in this PR?

There were some API changes after https://github.com/wireapp/wire-ios-transport/pull/127

Mainly because we dispatch all requests to a single session.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/127
- [x] https://github.com/wireapp/wire-ios-data-model/pull/653


vytis:
@zenkins release major

PR #108, opened 2019-01-25
Make sure to upload images on the background session

vytis:
### Issues

After some investigation it seems that we are uploading images on foreground session which means it competes for resources with normal messages.

### Solutions

Force the upload to happen from file url and this should make it happen on background session.


vytis:
@zenkins release patch

PR #114, opened 2019-02-18
Allow requests in background for asset client message

vytis:
### Issues

If file upload completes in the background the file is still showing as "in-progress" for the receiver.

### Causes

We would upload the file correctly, but would not update the message with the uploaded asset identifier. It was happening because the request strategy dies not allow requests in the background.

### Solutions

Allow requests in the background.


vytis:
@zenkins release patch

PR #122, opened 2019-03-13
ZIOS-11271: App freezes when loading image collections in Search UI

nicorsm:
## What's new in this PR?

App was freezing when loading the image collection in Search UI. This was caused by multiple download operations triggered while scrolling; seems to be partially solved with the refactoring of the last week.
I've added a notification after the completion of the download that triggers `loadImage()` of  `CollectionImageCell`, because cells weren't notified about that.


nicorsm:
@zenkins release patch

PR #124, opened 2019-03-14
Remove unused notifications

typfel:
## What's new in this PR?

There are no subscribers of the `.downloadFinishedNotificationName` and`.downloadFailedNotificationName` so I removed them.

Also added a test that we send a non-core-data change notification after downloading an asset.


typfel:
@zenkins release major

PR #125, opened 2019-03-20
Fix V2 asset request loop

typfel:
### Issues

Downloading a broken V2 asset would trigger a request loop

### Causes

Once a asset is whitelisted for downloading we will attempt to download over and over again until it succeeds, which it never happens since it's broken.

### Solutions

Check that `message.isDownloading` is true in the download filter predicate. We reset `isDownloading` to `false` when the request response arrives.


typfel:
@zenkins release patch

PR #126, opened 2019-03-21
ZIOS-11296: Add parsing mechanism for legacy link attachments

alexaubry:
### Issues

Due to compatibility issues between clients, we need to treat YouTube and SoundCloud links as a separate type of links, since they cannot be sent as link previews.

Their parsing was previously done in the UI and not persisted; which means we needed to recompute them for every rendering of the text message cells.

In addition to this performance bottleneck, we used the YouTube and SoundCloud API to manually fetch the details of the media, which this could have been achieved using the OpenGraph data provided by these websites at faster speeds and lower costs.

### Solutions

We add a new preprocessor for link attachments, that is based on the link preview preprocessor. It observes saves and requests the link attachment data from the link attachment detector.

We configure this preprocessor in the client message transcoder and use it as one of its `contextChangeTrackers`. 

### Dependencies

Needs release with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/701

### Notes

Using this branch, we will need to:
- update UI to opt-out of the old system (for now only for YouTube links)


Sources/Request Strategies/Assets/Helpers/LinkAttachmentsPreprocessor.swift
```diff
@@ -0,0 +1,134 @@
...
+            zmLog.debug("fetching previews for: \(message.nonce?.uuidString ?? "nil")")
+
+            // We DONT want to generate link previews inside a mentions
+            let mentionRanges = textMessageData.mentions.map(\.range)
+
+            // We DONT want to generate link previews for markdown links such as
+            // [click me!](www.example.com).
+            let markdownRanges = markdownLinkRanges(in: messageText)
+
+            linkAttachmentDetector.downloadLinkAttachments(inText: messageText, excluding: mentionRanges + markdownRanges) { [weak self] linkAttachments in
```
typfel: the code calculating the  excluded ranges are now duplicated in the `LinkPreviewPreprocessor`. Maybe it can be moved into shared a helper method?

alexaubry: I updated the PR in aea5e86 to use a base class for both the link attachment and preview preprocessors. The base has a generic argument we use to determine the result (either `LinkAttachment` or `LinkMetadata`).

alexaubry:
@zenkins release major

PR #128, opened 2019-03-26
Don't attempt download assets without .uploaded in the protobuf

typfel:
### Issues

App would crash after the ephemeral time runs out for an asset.

### Causes

We would try do download the asset after the content has been deleted which triggered a fatal assert.

### Solutions

Don't attempt to download assets which are missing the `.uploaded` protobuf.


typfel:
@zenkins release patch

PR #130, opened 2019-04-01
ZIOS-11330: Add blacklist property to link metadata

alexaubry:
### Issues

We do not want to display YouTube/SoundCloud/Giphy link previews in the UI, or make them behave like link previews (i.g. with removing the trailing link from the text).

### Solutions

Bump data model with auto exclusion of blacklisted link previews.


alexaubry:
@zenkins release minor

PR #135, opened 2019-04-23
Don't insert delivery confirmation when processing incoming message

typfel:
### Issues

We would send out multiple message delivery confirmations for incoming messages.

### Causes

We recently implemented support for confirming message delivery in batches after all events have been processed but didn't remove the old message delivery confirming code in the `ClientMessageTranscoder`.

### Solutions

Remove the now obsolete code for confirming message deliveries in `ClientMessageTranscoder` and update the tests.

## Notes

Since we no longer access any properties on  `ZMMessageUpdateResult in `ClientMessageTranscoder ` we simplify this API and return a ` ZMOTRMessage` . This will come in a separate PR.


typfel:
@zenkins release patch

PR #136, opened 2019-04-24
Update tests after re-factoring

typfel:
## What's new in this PR?

Update tests after re-factoring done in https://github.com/wireapp/wire-ios-data-model/pull/715

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/715


typfel:
@zenkins release major

PR #140, opened 2019-04-30
move libPhoneNumber from data model to sync engine

billypchan:
## What's new in this PR?

follow up of https://github.com/wireapp/wire-ios-sync-engine/pull/1016


billypchan:
@zenkins release major

PR #145, opened 2019-06-03
ZIOS-11451: Set legal hold status when sending message

alexaubry:
### Issues

We want to include a hint about legal hold presence when sending messages.

### Solutions

Update the client and asset message request strategies to insert the flag at sending time.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/731


alexaubry:
@zenkins release major

PR #146, opened 2019-06-03
Fetch details of newly discovered clients

typfel:
## What's new in this PR?

Since the security level now depends on the device class, we need fetch client details of a newly discovered client before we can update security level and decide if we should try to send message again.

- I've modified the `FetchingClientRequestStrategy` to fetch user client if the `needsToBeDownloadedFromBackend`is to true, which it is for new clients.

- I've introduced the `IdentifierObjectSync` as a generic  replacement for `ZMRemoteIdentifierObjectSync` since the UserClient remoteIdentifier is a string and not a UUID.

- A user client with `needsToBeDownloadedFromBackend`is to true is now considered a dependency for sending a message in conversation.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/730


Sources/Object Syncs/Helpers/IdentifierObjectSync.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+    func request(for identifiers: Set<T>) -> ZMTransportRequest?
+    
+    func didReceive(response: ZMTransportResponse, for identifiers: Set<T>)
+    
+}
+
+/// Class for syncing objects based on an identifier.
+
+public class IdentifierObjectSync<Transcoder: IdentifierObjectSyncTranscoder>: NSObject, ZMRequestGenerator {
```
vytis: Great approach here 👍

typfel:
@zenkins release major

PR #150, opened 2019-06-17
Fix: API changes for public logging

vytis:


vytis:
@zenkins release major

PR #152, opened 2019-06-20
Delete client & session if BE tells us it doesn't exist anymore

typfel:
### Issues

A message (like reaction) wouldn't get sent.

### Causes

The message was waiting some client metadata to be fetched before it could be sent. The fetching of the client metadata never completed because the backend responded with 404 since the clients had been deleted.

### Solutions

Delete clients when backend informs us that they no longer exists.


typfel:
@zenkins release patch

PR #156, opened 2019-06-28
Crash on verifying legal hold

typfel:
### Issues

The app would crash when force unwrapping the longer existing self client.

### Causes

The self client is never including the list of missing clients and was therefore deleted.

### Solutions

Exclude self clients when parsing the empty message response since we will receive these updates through other events.


typfel:
@zenkins release patch

PR #157, opened 2019-07-02
Fix missing edge case when verifying legal hold

typfel:
### Issues

When verifying a legal hold we send an empty message in the conversation to get an updated list of the active clients in the conversation. Because we send an empty message the response will always be a list of missing clients: 

```
missing: {
   user1: [ client1, client2]
   user2: [ client1]
}
```

This list represents the currently active user and clients in the conversation, we need to perform diff against our local state to figure which clients have been deleted or added. If a user is present locally but missing from the response it can mean two things:

1. User has been removed from the group and we are not aware of it
2. User has deleted all his/her clients 

We decided to always treat it as 2.

### Solutions

Delete all locally known clients when user is not appearing in the missing map.

### Todo
- [x] Implement suggested solution

### DIFF
--- a/Sources/Protocols/OTREntity.swift
+++ b/Sources/Protocols/OTREntity.swift
@@ -130,10 +130,15 @@ extension OTREntity {
 
         var changes: ZMConversationRemoteClientChangeSet = []
         var allMissingClients: Set<UserClient> = []
-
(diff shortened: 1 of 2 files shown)


typfel:
@zenkins release patch

PR #161, opened 2019-07-24
bump frameworks

billypchan:
## What's new in this PR?

bump frameworks for iOS13 crash fix


billypchan:
@zenkins release major

PR #163, opened 2019-08-15
Fix: Update dependencies in wire-ios-request-strategy

billypchan:
## What's new in this PR?

Fix the compile error due to change of method `addParticipantIfMissing` from https://github.com/wireapp/wire-ios-data-model/pull/768


WireRequestStrategy.xcodeproj/project.pbxproj
```diff
@@ -1078,10 +1078,9 @@
 			};
 			buildConfigurationList = 166901641D707509000FE4AF /* Build configuration list for PBXProject "WireRequestStrategy" */;
 			compatibilityVersion = "Xcode 3.2";
-			developmentRegion = English;
```
typfel: Revert

billypchan: It can be included in this PR, to fix the warning raised from XCode 10.2

typfel:
Tests also fail to build?

billypchan:
@zenkins release patch

PR #166, opened 2019-09-05
Bump components

typfel:
Due to major release of a dependency


typfel:
@zenkins release major

PR #167, opened 2019-09-26
Bump Data model

billypchan:
## What's new in this PR?

Bump Data model


billypchan:
@zenkins release minor

PR #168, opened 2019-09-26
Fix: move override method to data model

billypchan:
## What's new in this PR?

Move override method to Data model

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/786


billypchan:
@zenkins release minor

PR #169, opened 2019-09-27
bump for data model update

billypchan:
## What's new in this PR?

bump for data model update


billypchan:
@zenkins release major

PR #172, opened 2019-11-19
[Feature] Status Propagation

David-Henner:
## What's new in this PR?

Part of the status propagation feature.
Updated the filter predicate for the availability request strategy to only return a modified self user if the optimal limit of members in its team hasn't been reached.
This will ensure that status updates are propagated only when the team's member are below the optimal limit.

## Dependencies

https://github.com/wireapp/wire-ios-data-model/pull/806

## Notes

Will Update Cartfile once dependencies are released


zenkins:
Can one of the admins verify this patch?

billypchan:
@zenkins release major

PR #197, opened 2020-02-18
[Refactor] Generic message request strategy

David-Henner:
## What's new in this PR?

Use `GenericMessage` in Request Strategy

## Dependencies

https://github.com/wireapp/wire-ios-data-model/pull/895


typfel:
Seems like the tests doesn't compile.

Sources/Request Strategies/Generic Message/GenericMessageRequestStrategy.swift
```diff
@@ -72,15 +72,15 @@ public func ==(lhs: GenericMessageEntity, rhs: GenericMessageEntity) -> Bool {
...
     
     public func encryptedMessagePayloadData() -> (data: Data, strategy: MissingClientsStrategy)? {
-        return message.encryptedMessagePayloadData(conversation!, externalData: nil)
+        return message.zmMessage?.encryptedMessagePayloadData(conversation!, externalData: nil)
     }
-    
+
     public var debugInfo: String {
-        if message.hasCalling() {
+        if case .calling? = message.content {
```
johnxnguyen: What does the optional case mean? I'm guessing `content` is optional, but why can't we write `case .calling = message.content`?

David-Henner: The `switch / if case` pattern matches against the exact value of the object it evaluates. 
In `if case .calling? = message.content`, `message.content` is optional so we need to match against the optional `.calling?`.

It would be similar to doing `if let content = message.content, case .calling = content` (which doesn't compile)
Or, alternatively, `if case .some(.calling) = message.content`, which would work.

David-Henner:
@zenkins release major

PR #200, opened 2020-02-24
[Fix] Encryption error safe logging

David-Henner:
## What's new in this PR?

- Replaced error logs by safe public logs for the encryption errors

## Dependencies

https://github.com/wireapp/wire-ios-data-model/pull/899
https://github.com/wireapp/wire-ios-cryptobox/pull/36


Sources/Helpers/EncryptionSessionDirectory+UpdateEvents.swift
```diff
@@ -81,7 +81,8 @@ extension EncryptionSessionsDirectory {
     
     /// Appends a system message for a failed decryption
     fileprivate func appendFailedToDecryptMessage(after error: CBoxResult?, for event: ZMUpdateEvent, sender: UserClient, in moc: NSManagedObjectContext) {
-        zmLog.error("Failed to decrypt message with error: \(String(describing: error)), client id \(sender.remoteIdentifier!), event debug: \(event.debugInformation)")
+        zmLog.safePublic("Failed to decrypt message with error: \(EncryptionSessionDirectorySafeLog.cboxError(error)), client id \(EncryptionSessionDirectorySafeLog.senderIdentifier(sender))")
+        zmLog.error("event debug: \(event.debugInformation)")
```
David-Henner: Chose not to include the `event.debugInformation` as it's value could change in the future and potentially break the safety of public logs

Sources/Helpers/EncryptionSessionDirectory+UpdateEvents.swift
```diff
@@ -144,6 +145,23 @@ extension EncryptionSessionsDirectory {
...
+    case senderIdentifier(_ sender: UserClient)
+}
+
+extension EncryptionSessionDirectorySafeLog: SafeForLoggingStringConvertible {
+    public var safeForLoggingDescription: String {
+        switch self {
+        case .cboxError(let error):
+            return String(describing: error)
+        case .senderIdentifier(let userClient):
+            return "<\(userClient.remoteIdentifier!.readableHash)>"
```
billypchan: force unwrapped should be prevented

billypchan:
```suggestion
            return "<\(userClient.remoteIdentifier?.readableHash ?? "nil")>"
```

Sources/Helpers/EncryptionSessionDirectory+UpdateEvents.swift
```diff
@@ -81,7 +81,8 @@ extension EncryptionSessionsDirectory {
     
     /// Appends a system message for a failed decryption
     fileprivate func appendFailedToDecryptMessage(after error: CBoxResult?, for event: ZMUpdateEvent, sender: UserClient, in moc: NSManagedObjectContext) {
-        zmLog.error("Failed to decrypt message with error: \(String(describing: error)), client id \(sender.remoteIdentifier!), event debug: \(event.debugInformation)")
+        zmLog.safePublic("Failed to decrypt message with error: \(EncryptionSessionDirectorySafeLog.cboxError(error)), client id \(EncryptionSessionDirectorySafeLog.senderIdentifier(sender))")
```
typfel: UUID already conforms to `SafeForLoggingStringConvertible` in  https://github.com/wireapp/wire-ios-sync-engine/blob/ee322ceb76d55e1a9f511a0c9b65b29efe677b89/Source/Utility/UUID%2BSafeLogging.swift

Maybe it could be moved up and used here.

Sources/Helpers/EncryptionSessionDirectory+UpdateEvents.swift
```diff
@@ -144,6 +145,23 @@ extension EncryptionSessionsDirectory {
...
+// MARK: - Safe logging
+private enum EncryptionSessionDirectorySafeLog {
+    case cboxError(_ error: CBoxResult?)
+    case senderIdentifier(_ sender: UserClient)
+}
+
+extension EncryptionSessionDirectorySafeLog: SafeForLoggingStringConvertible {
+    public var safeForLoggingDescription: String {
+        switch self {
+        case .cboxError(let error):
```
typfel: Couldn't we confirm `CBoxError` to `SafeForLoggingStringConvertible`?

David-Henner: `sender.remoteIdentifier` is a `String`.
We could achieve the suggested changes by converting it to `UUID`
The `UUID` extension conforming to `SafeForLoggingStringConvertible` will then return  `transportString().readableHash`

Regarding `CBoxResult`, it can be conformed to `SafeForLogginStringConvertible` 

The following example shows how it would look like
```
extension CBoxResult: SafeForLoggingStringConvertible {
    public var safeForLoggingDescription: String {
        return String(describing: self)
    }
}

if let uuid = sender.remoteIdentifier {
    zmLog.safePublic("Failed to decrypt message with error: \(error), client id <\(uuid)>")
}
```

I created the enum `EncryptionSessionDirectorySafeLog` as an attempt to avoid converting `String` to `UUID` then back to `String`.

typfel: My issue about this enum is that it only solves the problem for this specific case. What about confirming UserClient to SafeForLogginStringConvertible?

David-Henner: `UserClient` inherits from `NSManagedObject` which itself is extended to conform to `SafeForLogginStringConvertible`. 

```
extension NSManagedObject: SafeForLoggingStringConvertible {
    public var safeForLoggingDescription: String {
        let moc: String = self.managedObjectContext?.description ?? "nil"
        
        return "\(type(of: self)) \(Unmanaged.passUnretained(self).toOpaque()): moc=\(moc) objectID=\(self.objectID)"
    }
}
```

However, in our case we only want to log `UserClient.remoteIdentifier`

David-Henner:
@zenkins release patch

PR #205, opened 2020-03-12
[Fix] Retry download asset when Cloud server error is 403.

marcomaddy1982:
### Issues

This PR solve the following edge case: when the back end redirects to the cloud service to get the image asset in a conversation, it could happen that the network bandwidth of the device is really rare. If the time interval is pretty long before the connectivity returns good the cloud responds with an error having status code 403

### Solutions

retry the image request and not delete the asset client message in case of status code 403.

### DIFF
--- a/Sources/Request Strategies/Assets/AssetV3DownloadRequestStrategy.swift
+++ b/Sources/Request Strategies/Assets/AssetV3DownloadRequestStrategy.swift
@@ -107,7 +107,7 @@ fileprivate let zmLog = ZMSLog(tag: "Asset V3")
         if response.result == .success {
             downloadSuccess = storeAndDecrypt(data: response.rawData!, for: assetClientMessage)
         }
-        // When the beck end redirects to the cloud service to get the image, it could be that the network bandwidth of the device is really rare. If the time interval is pretty long before the connectivity returns good the cloud responds with an error having status code 403 -> retry the image request and not delete the asset client message.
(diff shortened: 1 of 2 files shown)


zenkins:
Can one of the admins verify this patch?

Sources/Request Strategies/Assets/AssetV3DownloadRequestStrategy.swift
```diff
@@ -107,7 +107,8 @@ fileprivate let zmLog = ZMSLog(tag: "Asset V3")
         if response.result == .success {
             downloadSuccess = storeAndDecrypt(data: response.rawData!, for: assetClientMessage)
         }
-        else if response.result == .permanentError {
+        // When the beck end redirects to the cloud service to get the image, it could be that the network bandwidth of the device is really rare. If the time interval is pretty long before the connectivity returns good the cloud responds with an error having status code 403 -> retry the image request and not delete the asset client message.
```
typfel:
```suggestion
        // When the backend redirects to the cloud service to get the image, it could be that the network bandwidth of the device is really bad. If the time interval is pretty long before the connectivity returns, the cloud responds with an error having status code 403 -> retry the image request and do not delete the asset client message.
```

typfel: Maybe put in some line breaks to make the comment easier to read.

marcomaddy1982: beckend ahahaha. Sorry.

marcomaddy1982:
@zenkins release minor

marcomaddy1982:
@zenkins release minor

PR #209, opened 2020-04-01
[Feature] Limit status broadcast recipients

johnxnguyen:
### Issues

Broadcasting status messages to the entire team is not feasible for unbounded teams. Therefore we must limit the number of recipients who will receive status updates from a member.

### Solutions

Apply the changes introduce in https://github.com/wireapp/wire-ios-data-model/pull/939

## Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/939

### Attachments

**JIRA:** https://wearezeta.atlassian.net/browse/ZIOS-12827


Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
```diff
@@ -125,13 +133,14 @@ extension AvailabilityRequestStrategy : OTREntity {
     public func detectedMissingClient(for user: ZMUser) {
         // If we don't know about a user for a missing client we are out sync. To recover
         // from this we must restart the slow sync.
-        if !ZMUser.connectionsAndTeamMembers(in: managedObjectContext).contains(user) {
-            applicationStatus?.requestSlowSync()
+        if !ZMUser.recipientsForBroadcast(in: managedObjectContext, maxCount: maximumBroadcastRecipients).contains(user) {
+          applicationStatus?.requestSlowSync()
```
johnxnguyen, line 137: @typfel I'm not sure about this... do we actually only want to request a slow sync if the user with a missing client is supposed to be a recipient?

typfel: This will never happen now that we don't broadcast to everyone. The backend normally ensures that we encrypt the message to all clients, but now that we only broadcast to a subset of the users we explicitly tell the backend which users we want the backend ensure that we have encrypted payloads for. Since we will only include users that we know about this scenario shouldn't happen.

johnxnguyen: But couldn't it happen that one of our recipients has a new client that we don't know about? In which case we would want to invert the condition here to:

```swift
if ZMUser.recipientsForBroadcast(in: ..., maxCount: ...).contains(user) {
  applicationStatus?.requestSlowSync()
}
```

johnxnguyen:
@David-Henner no, but it'd be nice to get it in the projects :)

Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
```diff
@@ -50,21 +52,27 @@ public class AvailabilityRequestStrategy : AbstractRequestStrategy {
...
+        guard let dataAndMissingClientStrategy = message.encryptedMessagePayloadDataForBroadcast(recipients: recipients, in: context) else {
             return nil
         }
-        
+
         let protobufContentType = "application/x-protobuf"
         let path = originalPath.pathWithMissingClientStrategy(strategy: dataAndMissingClientStrategy.strategy)
-        let request = ZMTransportRequest(path: path, method: .methodPOST, binaryData: dataAndMissingClientStrategy.data, type: protobufContentType, contentDisposition: nil)
+
+        let request = ZMTransportRequest(path: path,
```
typfel, line 71: We need to tell the backend not to report the missing users we didn't include. See https://github.com/zinfra/backend-issues/issues/1248

johnxnguyen: We do, it's in [data model](https://github.com/wireapp/wire-ios-data-model/pull/939/files#diff-4868997c91e36837178bb65c0ec40c5bR119)

Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
typfel: Yes that can still happen but we don't need to start a slow sync for a missing client. A missing client is not an indication that we are out of sync.

Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
typfel: Ah 👍

Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
johnxnguyen: Ok cool, I'll delete this code then.

Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
```diff
@@ -131,11 +131,8 @@ extension AvailabilityRequestStrategy: OTREntity {
     }
     
     public func detectedMissingClient(for user: ZMUser) {
-        // If we don't know about a user for a missing client we are out sync. To recover
-        // from this we must restart the slow sync.
-        if !ZMUser.recipientsForBroadcast(in: managedObjectContext, maxCount: maximumBroadcastRecipients).contains(user) {
-          applicationStatus?.requestSlowSync()
-        }
+        // Broadcast messsages are targeted to specific set of recipients and we will ignore any other
+        // users who are not recipients. This should never be called.
```
typfel, line 135:
```suggestion
        // users who are not recipients.
```

This will be called when we miss a client, but we don't need to react on it.

PR #214, opened 2020-04-15
[Feature]  Asset decryption error fix

marcomaddy1982:
### Issues

when the client was not able to decrypt an asset a request was done in loop.

### Solutions

when the asset could not be decrypted it will be deleted from the conversation.


typfel:
Let's add unit test for this scenario in `AssetV3DownloadRequestStrategyTests`

PR #218, opened 2020-04-24
[Fix] Don't mark clients with established sessions as missed

typfel:
### Issues

We had an incident (internal) were encryption sessions broke, which resulted in decryption errors.

### Causes

The session broke because the backend unexpectedly reported client as missing when they weren't due to mismatch between backend and client rollout. We the client was blindly trusting the server that client were missing and re-fetched the pre-keys and established new sessions, effectively breaking the sessions.

Even though this shouldn't happen when the client and backend is in sync but it still would  be nice to be resilient towards these kind of issues.

### Solutions

Verify that a client doesn't already have an established session before adding it to the list of missing clients.


typfel:
@zenkins release patch

PR #220, opened 2020-04-27
[Feature] re-fetch redundant users in order to discover if they are deleted

typfel:
## What's new in this PR?

We already re-fetch user metadata when we discover that the last client of user is deleted when sending a message. In addition to that we also need to re-fetch the metadata when a user is reported as redundant when sending a message. 

Being redundant means that the user is no longer part of the conversation, which could mean that the user left the conversation and you missed the leave event or that the user has deleted his account and you missed the user deletion event. In either case we resolve it by re-fetching the conversation and the user metadata.

## Notes

There's quite some code duplication going on here but fixing it would be too big of a task include in this PR. I'll create a separate re-factoring ticket for this.

### DIFF
--- a/Sources/Helpers/ZMMessage+Dependency.swift
+++ b/Sources/Helpers/ZMMessage+Dependency.swift
@@ -37,11 +37,15 @@ extension ZMOTRMessage: OTREntity {
         return dependent ?? super.dependentObjectNeedingUpdateBeforeProcessing
     }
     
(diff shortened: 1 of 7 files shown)


typfel:
Re-factoring ticket: https://wearezeta.atlassian.net/browse/ZIOS-13181

typfel:
@zenkins release major

PR #221, opened 2020-04-29
[Refactor] Remove code duplication when processing message sending responses

typfel:
## What's new in this PR?

This is a follow up on https://github.com/wireapp/wire-ios-request-strategy/pull/220 to remove some code duplication.

### Issues

Parts of the business logic for handling redundant and missing clients was duplicated in all the objects implementing the `OTREntity` protocol

### Solutions

Add an optional `conversation: ZMConversation?` property to the`OTREntity` protocol and move the duplicated business logic into the protocol methods for parsing redundant and missing clients.


typfel:
@zenkins release major

PR #224, opened 2020-05-15
[Feature] Client discovery strategy

johnxnguyen:
## What's new in this PR?

This PR introduces the `ClientDiscoveryRequestStrategy`, a strategy designed to fetch a list of all clients in a given conversation. This strategy is intended to be used to provide AVS with a client list for use with SFT calls.

It works by sending an OTR message with an empty payload and zero recipients in a specific conversation. The backend will return a 412 response indicating that the message is missing recipients (as expected). In the payload of the response we will find a list of all missing clients, which will be the list of all clients in the conversation. This list is then passed into a completion handler.

### DIFF
--- a/Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
+++ b/Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
@@ -24,7 +24,7 @@ import Foundation
 public final class ClientDiscoveryRequestStrategy: NSObject, RequestStrategy {
 
   public typealias ClientIdsByUserId = [String: [String]]
-  public typealias ClientDiscoveryRequestCompletion = (ClientIdsByUserId) -> Void
+  public typealias RequestCompletion = (ClientIdsByUserId) -> Void
 
   // MARK: - Private properties
 
@@ -34,7 +34,7 @@ public final class ClientDiscoveryRequestStrategy: NSObject, RequestStrategy {
   private var requestSync: ZMSingleRequestSync!
 
   private var pendingRequest: ZMTransportRequest?
-  private var pendingCompletion: ClientDiscoveryRequestCompletion?
+  private var pendingCompletion: RequestCompletion?
 
   // MARK: - Init
 
@@ -56,7 +56,7 @@ public final class ClientDiscoveryRequestStrategy: NSObject, RequestStrategy {
   ///   - conversationId: the id of the conversation.
   ///   - completion: invoked with a map of user ids to an array of client ids.
 
-  public func requestClientList(conversationId: UUID, completion: @escaping ClientDiscoveryRequestCompletion) {
+  public func requestClientList(conversationId: UUID, completion: @escaping RequestCompletion) {
     managedObjectContext.performGroupedBlock { [unowned self] in
       guard let selfClient = ZMUser.selfUser(in: self.managedObjectContext).selfClient() else { return }
 
--- a/Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
+++ b/Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
@@ -1,0 +1,92 @@
+// Wire
+// Copyright (C) 2019 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
```diff
@@ -0,0 +1,92 @@
...
+    return requestSync.nextRequest()
+  }
+
+  /// Triggers a request to fetch the list of clients in a given conversation.
+  ///
+  /// - Parameters:
+  ///   - conversationId: the id of the conversation.
+  ///   - completion: invoked with a map of user ids to an array of client ids.
+
+  public func requestClientList(conversationId: UUID, completion: @escaping RequestCompletion) {
```
typfel, line 59: What happens if this is called twice? Only one of the request will have it's completion handler called?

Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
```diff
@@ -0,0 +1,92 @@
...
+import Foundation
+
+/// This strategy is used to get an up-to-date list of all clients in a given conversation.
+/// It achieves this by posting an empty OTR message with no recipients in the coversation,
+/// then parsing the 412 error response to discover the list of missing clients reported by
+/// the backend.
+
+public final class ClientDiscoveryRequestStrategy: NSObject, RequestStrategy {
+
+  public typealias ClientIdsByUserId = [String: [String]]
```
typfel, line 26: Would it make sense to define `Codable` structs for this?

Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
johnxnguyen: Yes, only the latest handler will be called, but it would only be called once since we "consume" the handler once it is invoked. I was modeling this behavior off the `CallingRequestStrategy` in sync engine, in particular how we request the call config.

typfel: As user of this generically named strategy I would expect a second call to cancel the first request (maybe it's called difference places in the code). Couldn't we keep a list of pending completion handlers?

Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
johnxnguyen: Hmm... I'm not sure it does make sense, because the keys contain data (the userId). We could do it but it would require an explicit decodable implementation.

Also... can you decode a dictionary? We don't have the JSON string here but an instance `ZMTransportData`.

typfel: I would suggest to create a struct for the whole response with this part being a nested struct.

UserID is a UUID which already implements `Codable` so it should work out of the box.

typfel: We could then re-use those those structs in other strategies, but I understand if you don't want to do this right now if you are under time pressure.

Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
typfel: Example of nested payload structs https://github.com/wireapp/wire-ios-sync-engine/blob/a9dd002607a85bb39ca5df2511f071cb8f0eea97/Source/Synchronization/Strategies/PermissionsDownloadRequestStrategy.swift#L22

Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
typfel: Example of nested payload structs https://github.com/wireapp/wire-ios-sync-engine/blob/a9dd002607a85bb39ca5df2511f071cb8f0eea97/Source/Synchronization/Strategies/PermissionsDownloadRequestStrategy.swift#L22

typfel: It's probably better to do it as a separate PR.

johnxnguyen: Noting for future self: `ZMTransportResponse` has a `rawData` property which contains the JSON payload as a blob.

Sources/Request Strategies/User Clients/ClientDiscoveryRequestStrategy.swift
johnxnguyen: Sure, I'll give it a go.

johnxnguyen: So, it's using a `ZMSingleRequestSync`, meaning we only process one request at a time. For the use case with AVS, it seemed like an appropriate choice. The way that I would use it is to make it be owned by the `CallingRequestStrategy` (not the `ZMSyncStrategy`) and use it there. What do you think?

johnxnguyen:
Implemented in https://github.com/wireapp/wire-ios-sync-engine/pull/1235

PR #226, opened 2020-05-18
[Refactor] Change process method in PushMessageHandler protocol

KaterinaWire:
## What's new in this PR?

Modified process() method in PushMessageHandler protocol, because this protocol must be used for objc code.


Sources/Protocols/PushMessageHandler.swift
```diff
@@ -36,5 +37,6 @@ public protocol PushMessageHandler: NSObjectProtocol {
     /// Shows a notification for a failure to send
     ///
     /// - Parameter message: message that failed to send
+    @objc(processMessage:)
```
typfel, line 40: This `@objc` is on the wrong method?

KaterinaWire: Fixed

PR #230, opened 2020-06-10
[Feature] Targeted messages

johnxnguyen:
## What's new in this PR?

The upcoming conference calling feature requires clients to be able to send targeted encrypted messages to specific client devices. This PR includes changes to make this possible.

### Changes

API changes introduce in https://github.com/wireapp/wire-ios-data-model/pull/993 such as method renaming have been addressed.

Additionally, when scheduling a generic message, there is now the option to specify the target recipients. There are two cases:

-  `conversationParticipants` (the default), which preserves the existing behavior of sending a message to all the participants of the conversation,
- `clients`, which contains a collection of specific client devices which should be the **only** recipients of the message.

## Dependencies
Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/993


johnxnguyen:
@zenkins release major

PR #234, opened 2020-07-29
[Fix] ignore_missing query parameter has no value

johnxnguyen:
### Issues

When sending a targeted message, the backend response with a 412 error stating there are missing clients.

### Causes

The query parameter `ignore_missing` has no value specified.

### Solutions

Set the value to the parameter to `true`


johnxnguyen:
@zenkins release patch

PR #235, opened 2020-07-29
[Feature] Add strategy for sending delivery receipts

typfel:
## What's new in this PR?

Add `DeliveryReceiptRequestStrategy` which sends out delivery receipts by processing received update events and without relying on that messages we are sending receipts exists in the database. This change is necessary since we will not be able to insert message in the background when EAR is enabled but we still want to send out delivery receipts.

Remove the `DeliveryConfirmationDelegate` which used in the `ClientMessageTranscoder` to notify when delivery receipts were sent. This is not necessary anymore now that delivery receipts are sent using a dedicated request strategy.

Update after `ZMUpdateEvent` accessor methods were converted to properties so that they can we accessed via key paths.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1006

## Notes

As future re-factoring it would be nice to inject the `GenericMessageRequestStrategy` and maybe put it behind a protocol, this would make testing these kind of strategies much easier. For now I exposed a method as internal to achieve a similar goal.


Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
```diff
@@ -0,0 +1,139 @@
...
+        
+        genericMessageStrategy.schedule(message: GenericMessage(content: confirmation),
+                                        inConversation: deliveryReceipt.conversation,
+                                        targetRecipients: .users(Set(arrayLiteral: deliveryReceipt.sender)),
+                                        completionHandler: nil)
+        
+    }
+    
+    func deliveryReceipts(for events: [ZMUpdateEvent]) -> [DeliveryReceipt] {
+        let messageByConversation = events.filter { (event) -> Bool in
```
johnxnguyen, line 110: Suggestion: `messagesByConversation` or `eventsByConversation`

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
```diff
@@ -0,0 +1,139 @@
...
+            return event.type.isOne(of: .conversationOtrMessageAdd, .conversationOtrAssetAdd)
+        }.partition(by: \.conversationID)
+        
+        var deliveryReceipts: [DeliveryReceipt] = []
+        
+        messageByConversation.forEach { (conversationID: UUID, events: [ZMUpdateEvent]) in
+            guard let conversation = ZMConversation.fetch(withRemoteIdentifier: conversationID,
+                                                          in: managedObjectContext) else { return }
+            
+            let messagesBySender = events
```
johnxnguyen, line 120: Suggestion: `eventsBySender`

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategyTests.swift
```diff
@@ -0,0 +1,195 @@
...
+            // given
+            let events = [self.createTextUpdateEvent(from: self.otherUser, in: self.oneToOneConversation),
+                          self.createTextUpdateEvent(from: self.otherUser, in: self.secondOneToOneConveration)]
+            
+            // when
+            let deliveryReceipts = self.sut.deliveryReceipts(for: events)
+            
+            // then
+            XCTAssertEqual(deliveryReceipts.count, 2)
+            XCTAssertEqual(deliveryReceipts.first?.messageIDs.count, 1)
```
johnxnguyen, line 154: Do you also need to assert the second delivery receipt?

PR #237, opened 2020-08-17
[Feature] Encryption at rest

typfel:
## What's new in this PR?

Add support for encrypting the message data when its persisted to the database which can only be decrypted after the user authenticates inside the application.

---

- Add `DeliveryReceiptRequestStrategy` for sending delivery receipts for incoming messages
- Decouple the ClientMessageTranscoder from the message notification processing

## Dependencies
Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1014


johnxnguyen:
@zenkins release major

PR #238, opened 2020-08-19
[Fix] Don't send delivery receipts for confirmation messages

johnxnguyen:
### Issues

The app is continuously sending delivery receipts.

### Causes

We are sending delivery receipts for confirmation messages (aka delivery receipts), so if we send a delivery receipt, the other client will send a receipt for that receipt, for which we will send a receipt for that receipt...

### Solutions

Don't send receipts for confirmation messages.

### DIFF
--- a/Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
+++ b/Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
@@ -24,6 +24,7 @@ extension ZMUpdateEvent {
         
     func needsDeliveryConfirmation(_ currentDate: Date = Date(),
                                    managedObjectContext: NSManagedObjectContext) -> Bool {
+
         guard
             let conversationID = conversationUUID,
             let conversation = ZMConversation.fetch(withRemoteIdentifier: conversationID, in: managedObjectContext), conversation.conversationType == .oneOnOne,
@@ -99,10 +100,10 @@ extension DeliveryReceiptRequestStrategy: ZMEventConsumer {
     }
     
     func deliveryReceipts(for events: [ZMUpdateEvent]) -> [DeliveryReceipt] {
-        let eventsByConversation = events.filter { (event) -> Bool in
-            return event.type.isOne(of: .conversationOtrMessageAdd, .conversationOtrAssetAdd)
-        }.partition(by: \.conversationUUID)
-        
+        let eventsByConversation = events
+            .filter { GenericMessage(from: $0)?.requiresDeliveryReceipt ?? false }
+            .partition(by: \.conversationUUID)
+
         var deliveryReceipts: [DeliveryReceipt] = []
         
         eventsByConversation.forEach { (conversationID: UUID, events: [ZMUpdateEvent]) in
@@ -128,4 +129,12 @@ extension DeliveryReceiptRequestStrategy: ZMEventConsumer {
         return deliveryReceipts
     }
     
+}
+
+private extension GenericMessage {
+
+    var requiresDeliveryReceipt: Bool {
+        return !hasConfirmation
+    }
+
 }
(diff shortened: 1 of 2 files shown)


Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategyTests.swift
```diff
@@ -173,24 +187,36 @@ class DeliveryReceiptRequestStrategyTests: MessagingTestBase {
...
     
-    func createTextUpdateEvent(from: ZMUser,
-                               in conversation: ZMConversation,
-                               timestamp: Date = Date()) -> ZMUpdateEvent {
-        let conversationID = conversation.remoteIdentifier!.transportString()
+    func createTextUpdateEvent(
+        from sender: ZMUser,
+        in conversation: ZMConversation,
+        timestamp: Date = Date()
+    ) -> ZMUpdateEvent {
```
billypchan, line 194: We should keep the original indent.

billypchan (approved):

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategyTests.swift
johnxnguyen: Done

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
```diff
@@ -99,10 +100,10 @@ extension DeliveryReceiptRequestStrategy: ZMEventConsumer {
     }
     
     func deliveryReceipts(for events: [ZMUpdateEvent]) -> [DeliveryReceipt] {
-        let eventsByConversation = events.filter { (event) -> Bool in
-            return event.type.isOne(of: .conversationOtrMessageAdd, .conversationOtrAssetAdd)
-        }.partition(by: \.conversationUUID)
-        
+        let eventsByConversation = events
+            .filter { GenericMessage(from: $0)?.requiresDeliveryReceipt ?? false }
```
typfel, line 104: Why not add this as a condition inside `needsDeliveryConfirmation()`?

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
```diff
@@ -129,3 +130,11 @@ extension DeliveryReceiptRequestStrategy: ZMEventConsumer {
     }
     
 }
+
+private extension GenericMessage {
+
+    var requiresDeliveryReceipt: Bool {
+        return !hasConfirmation
```
typfel, line 137: Was trying to figure out how this was done before I can't find any filtering based not the generic message, but it seems like we should not send delivery receipts for many types:

Reaction, Calling, Availability, ClientAction, LastRead, Cleared, MessageHide, MessageDelete, ButtonAction, ButtonActionConfirmation


or inverted we should only send for:

Text, ImageAsset, Knock, External, Asset, Location, Ephemeral, Composite

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
johnxnguyen: I thought maybe it's better to filter out early to avoid iterating over unnecessary events. What do you think?

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
johnxnguyen: It's best to be explicit so I will make the condition more complete.

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
typfel: I think it's not expensive to iterate over the events.

johnxnguyen: Moved

Sources/Request Strategies/Delivery Receipts/DeliveryReceiptRequestStrategy.swift
johnxnguyen: Done

typfel (approved):

typfel:
@zenkins release patch

PR #239, opened 2020-08-24
[Fix] Avoid sending encrypted messages while fetching the notification stream

typfel:
### Issues

It was possible that the `DeliveryReceiptRequestStrategy` could start sending encrypted messages before we've finished fetching and decrypting all encrypted events in the notification stream.

### Causes

The `DeliveryReceiptRequestStrategy` was not a subclass of `AbstractRequestStrategy` and therefore weren't following any rules about when it's allowed to make requests.

### Solutions

Make `DeliveryReceiptRequestStrategy` an `AbstractRequestStrategy` and configure it not run during quickSync phase, which is the phase where we download and decrypt events.

In order to express this the `ZMStrategyConfigurationOption` was expanded to differentiate between: 
- slow syncing
- quick syncing 
- waiting for the web socket  connection.

### DIFF
--- a/Sources/Request Strategies/Base Strategies/AbstractRequestStrategyTests.swift
+++ b/Sources/Request Strategies/Base Strategies/AbstractRequestStrategyTests.swift
@@ -67,29 +67,147 @@ class AbstractRequestStrategyTests : MessagingTestBase {
     let mockApplicationStatus = MockApplicationStatus()
     
     func checkAllPermutations(on sut : RequestStrategy & TestableAbstractRequestStrategy) {
-        
-        assertPass(withConfiguration: [.allowsRequestsWhileOnline], operationState: .foreground, synchronizationState: .online, sut: sut)
-        assertPass(withConfiguration: [.allowsRequestsDuringSlowSync], operationState: .foreground, synchronizationState: .slowSyncing, sut: sut)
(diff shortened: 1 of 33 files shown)


Sources/Request Strategies/Base Strategies/AbstractRequestStrategyTests.swift
```diff
@@ -67,45 +67,171 @@ class AbstractRequestStrategyTests : MessagingTestBase {
...
+
+        assertFail(withConfiguration: [.allowsRequestsWhileOnline],
+                   operationState: .foreground,
+                   synchronizationState: .quickSyncing,
+                   sut: sut)
+
+        assertFail(withConfiguration: [.allowsRequestsWhileOnline],
+                   operationState: .foreground,
+                   synchronizationState: .unauthenticated,
+                   sut: sut)
```
johnxnguyen, line 96: The configuration specifies exactly during which synchronization states a request can be made, is this correct?

johnxnguyen (approved):

Sources/Request Strategies/Base Strategies/AbstractRequestStrategyTests.swift
typfel: Yes exactly, it's not just synchronisation state though, it can be application state as well like foreground/background.

typfel:
@zenkins release major

PR #240, opened 2020-08-26
[Refactor] Move ZMLocalNotification class from SE

KaterinaWire:
## What's new in this PR?

To be able to convert an update event to a localNotification in `Notification engine`, we need to have `ZMLocalNotification` in the `Request strategy`. This PR contains ZMLocalNotification and all required classes are moved from SE.


Cartfile.resolved
```diff
@@ -3,11 +3,12 @@ github "wireapp/PINCache" "2.3-swift3.1"
...
-github "wireapp/wire-ios-data-model" "refactor/notification-extension"
+github "wireapp/wire-ios-data-model" "refactor/move-ZMKeyValueStore-from-SE"
 github "wireapp/wire-ios-images" "31.0.0"
 github "wireapp/wire-ios-link-preview" "27.0.1"
 github "wireapp/wire-ios-protos" "24.0.0"
 github "wireapp/wire-ios-system" "31.0.1"
 github "wireapp/wire-ios-testing" "22.0.0"
 github "wireapp/wire-ios-transport" "70.0.0"
 github "wireapp/wire-ios-utilities" "refactor/notification-extension"
+github "wireapp/avs-ios-binaries" "6.0.223"
```
typfel, line 14: Why did you add AVS as dependency?

typfel:
All test classes haven't been moved like: LocalNotificationDispatcherTests

Sources/Notifications/Push Notifications/LocalNotificationDispatcher.swift
```diff
@@ -0,0 +1,126 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UserNotifications
+
+/// Creates and cancels local notifications
+@objcMembers public class LocalNotificationDispatcher: NSObject {
```
typfel, line 23: Do you really need to move the `LocalNotificationDispatcher`?

billypchan:
We need to rewrite pipeline to download Crowdin string resource after this PR is merger.

Sources/Notifications/Push Notifications/LocalNotificationDispatcher.swift
KaterinaWire: No, I removed it.

Cartfile.resolved
KaterinaWire: Because I moved `CallClosedReason`

typfel (approved):

KaterinaWire:
> We need to rewrite pipeline to download Crowdin string resource after this PR is merger.

I will do it when I merge this PR to the develop branch.

Sources/Notifications/Push Notifications/Notification Types/LocalNotificationType+Configuration.swift
```diff
@@ -0,0 +1,102 @@
...
+        switch self {
+        case .calling(let callState):
+            switch (callState) {
+            case .incoming:
+                category = .incomingCall
+            case .terminating(reason: .timeout):
+                category = .missedCall
+            default :
+                category = PushNotificationCategory.conversation.addMuteIfNeeded(hasTeam: hasTeam)
+            }
```
johnxnguyen, line 52: This seems to be out of date with the latest sync engine release.

Cartfile.resolved
johnxnguyen: But do you need AVS or just the enum? I suggest to try to see if you can extract what you need for notifications away from the AVS library.

johnxnguyen (changes requested):
There have been some changes to this code on Sync Engine which aren't present here. Don't forget to include them.

Sources/Notifications/Push Notifications/Notification Types/LocalNotificationType+Configuration.swift
KaterinaWire: Updated

Cartfile.resolved
KaterinaWire: Right now yes. But the next step will be to handle call notifications in notification service extension, and we need to move more  classes to the request strategy related to the calling.

johnxnguyen: Which calling related classes do you need to move?

typfel: We will need figure out if we have a incoming call in the notification engine, it means we'll need to forward events to AVS and listen to their callbacks. Either we need to split that part out of the CallingRequestStrategy and WireCallCenter or build something new for the notification engine.

What do think @johnxnguyen ?

johnxnguyen: So we need to

- process a call message event, pass it to AVS
- AVS will invoke the incoming call handler
- `WireCallCenter` will broadcast a new incoming call state
- `CallStateObserver` responds to the state change and tells the `LocalNotificationDispatcher` to make the local notification.

It doesn't make sense to move **all** of the calling code to the request strategy framework just so we can figure out if we need to display one type of notification (which is only going to happen if CallKit is disabled).

I'm hesitant to interface with AVS over two different frameworks, but it would probably be fine if in the notification engine AVS was used in an absolutely minimal and isolated way.

So the notification engine sits next to sync engine and share engine? That is, between the UI and request strategy?

typfel: > So the notification engine sits next to sync engine and share engine? That is, between the UI and request strategy?

Exactly

> It doesn't make sense to move all of the calling code to the request strategy framework just so we can figure out if we need to display one type of notification (which is only going to happen if CallKit is disabled).

No, maybe it make sense to re-implement this use case in the notification engine.

johnxnguyen: If we re-implement this use case in the notification engine, then we would have two instances of AVS, one that would be run while the app is in the foreground, and the other run in the notification extension while the main app is in the background (is this correct?). This would mean that if we opened the app to answer the call, the foreground instance of AVS wouldn't know there was an incoming call.

typfel: As described here https://github.com/wireapp/ios-architecture/issues/90

The idea is that we will process the call events in notification extension in order figure out if there's an incoming call so that we can display a calling notification / notify CallKit. We are not deleting any events in the notification extension, that means we will process same calling events again when the main app is launched. That way AVS will be up to date on the call state.

This works because AVS is not persisting any call state, it's only kept in memory.

johnxnguyen: Great, thanks for explanation. I would see what this miniature call center would look like, if it's compact and concise then it would suitable. But it could be that it doesn't make so much sense in which case we should find another solution.

Already thinking about it, I can see that creating a new instance of AVS would require us to register a lot of callbacks that would never get used. This would lead to a lot of unnecessary duplication.

typfel: > Already thinking about it, I can see that creating a new instance of AVS would require us to register a lot of callbacks that would never get used. This would lead to a lot of unnecessary duplication.

Mm maybe the `AVSWrapper` could be shared if it's decoupled from the `WireCallCenter`

KaterinaWire: @johnxnguyen Could it be a temporary solution to move `CallClosedReason` and add AVS s a dependency to the request strategy? And I can add a ticket (as a reminder) to discuss/refactor the calling events.

johnxnguyen: Sure, but we should try to understand what the proper solution to this will be as it may require a bit of work.

PR #243, opened 2020-09-09
[Refactor] Add pushNotificationStatus to the NotificationStreamSync

KaterinaWire:
## What's new in this PR?

Add `pushNotificationStatus` property.  Fetch events if pushNotificationStatus.hasEventsToFetch.


typfel (changes requested):
I would suggest not add link the `PushNotificationStatus` to the `NotificationStreamSync`. The sync objects ideally be re-usable and not tied to the particular logic of any strategy.

Try to whatever logic you need directly to the `PushNotificationStrategy`.

PR #246, opened 2020-10-07
[Chore] Bump components and fix failing test

typfel:
### Issues

Tests were failing after bumping DM.

### Causes

We no longer calculate the unread count on `awakeFromFetch` this had the side effect that the message deletion in the test was not processed.

### Solutions

Manually trigger the processing by calling `processPendingChanges()`


billypchan (approved):

billypchan:
@zenkins release patch

PR #250, opened 2020-11-04
[Performance] Use more efficient method to check existence of data in cache

johnxnguyen:
### Issues

The predicate used to determine which `ZMClientMessage` instances require image processing is inefficient.

### Causes

The predicate checks there exists any data in the asset cache for the given message. It does this by trying to read the data (in its entirety) and checking if it exists. This relatively slow because it has to read data from disk.

### Solutions

Use a more appropriate method that checks for the existence of data without actually reading it all.


typfel (approved):

marcomaddy1982 (approved):

johnxnguyen:
@zenkins release patch

PR #254, opened 2020-11-17
[Feature] Define the basic interface for feature controller

KaterinaWire:
### Issues
Add a FeatureController that is responsible for :
- parse and save configs to the Core Data,
- notify UI (Applock) / features.

Needs releases with:

- https://github.com/wireapp/wire-ios-data-model/pull/1066
- https://github.com/wireapp/wire-ios-data-model/pull/1067

### Attachments
https://miro.com/app/board/o9J_lfiqql4=/


Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+//
+
+import Foundation
+
+public protocol Configurable {
+    associatedtype Config: Codable
+    static var name: String { get }
+}
+
+public enum FeatureModel {
```
johnxnguyen, line 26: I think we should find a more suitable name, but not quite sure what just yet.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+            private enum CodingKeys: String, CodingKey {
+                case enforceAppLock = "enforce_app_lock"
+                case inactivityTimeoutSecs = "inactivity_timeout_secs"
+            }
+        }
+    }
+}
+
+public struct FeatureConfigResponse<T: Configurable>: Decodable {
+    var status: String
```
johnxnguyen, line 42: The value of `status` is always expected to be `enabled` or `disabled`, right? In this case, we could define and use an enum:

```swift
enum Status: String, Codable {
  case enabled
  case disabled
}
```

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+    public init(managedObjectContext: NSManagedObjectContext) {
+        moc = managedObjectContext
+    }
+    
+    public func status<T: Configurable>(for feature: T.Type) -> Bool {
+        guard let feature = Feature.fetch(feature.name, context: moc) else {
+                return false
+        }
+        return Bool(statusStr: feature.status)
+    }
```
johnxnguyen, line 80: I suggest to use the `Status` enum I mentioned earlier as the return type because it would read better and make more sense. E.g what does this mean? `if status(for: AppLock.self) { ... }`

johnxnguyen: All features have a status, but not all features are configurable. The type constraint here only returns the status for configurable features. Not a major issue now since we only have one feature and it is configurable, but it would be a problem.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+    
+    public func configuration<T: Configurable>(for feature: T.Type) -> T.Config? {
+        guard let configData = Feature.fetch(feature.name, context: moc)?.config else {
+                return nil
+        }
+        return try? JSONDecoder().decode(feature.Config, from: configData)
```
johnxnguyen, line 86: I suggest to use the generic parameter directly instead of the argument, for readability sake. I.e `T.name` instead of `feature.name` and `T.Config` instead of `feature.Config`.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+
+// MARK: - Internal
```
johnxnguyen, line 91: This says internal but the methods are public.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+extension FeatureController {
+    public func save<T: Configurable>(_ feature: T.Type, data: Data) {
+        do {
+            let configuration = try JSONDecoder().decode(FeatureConfigResponse<FeatureModel.AppLock>.self, from: data)
```
johnxnguyen, line 95: The `feature` argument isn't being used, do you mean to write this?

```swift
try JSONDecoder().decode(FeatureConfigResponse<T>.self, from: data)
```

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+            let appLockFeature = Feature.createOrUpdate(feature.name,
+                                                        status: configuration.status,
+                                                        config: configuration.configData,
+                                                        context: moc)
+            moc.saveOrRollback()
+            
+            if let appLockFeature = appLockFeature  {
+                NotificationCenter.default.post(name: FeatureController.needsToUpdateFeatureNotificationName, object: nil, userInfo: ["appLock" : appLockFeature])
+            }
```
johnxnguyen, line 104: Are you planning to generalize this in a later PR? If so, add a todo comment with your name so we don't forget.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+            
+        } catch {}
```
johnxnguyen, line 106: We should in the very least log the failure so we can become aware of decoding issues.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,137 @@
...
+
+private extension Bool {
+    init(statusStr: String) {
+        switch statusStr {
+        case "enabled":
+            self = true
+        case "disabled":
+            self = false
+        default:
+            self = false
```
johnxnguyen, line 134: The `Status` enum would make this redundant.

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -0,0 +1,78 @@
...
+        let json: [String : Any] = ["applock": [
+            "status": "disabled",
+            "config": [
+            "enforce_app_lock": true,
+            "inactivity_timeout_secs": 30
+            ]
+            ]
+            ]
+        
+        let data = try? JSONSerialization.data(withJSONObject: json, options: JSONSerialization.WritingOptions.prettyPrinted)
```
johnxnguyen, line 48: You could also write the json object as a string (which is cleaner in my opinion):

```swift
let json = """
{
  "applock": {
      "status": "disabled",
      ...
    }
}
"""

let data = json.data(using: .utf8)!
```

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -0,0 +1,78 @@
...
+        
+        // When
+        sut.saveAllFeatures(data!)
+       
+        // Then
+        let featureStatus = sut.status(for: FeatureModel.AppLock.self)
```
johnxnguyen, line 54: Since we're testing the save method, we should check the database directly instead of fetching the status. Testing the status method should be done in a separate test.

Sources/Helpers/FeatureController.swift
KaterinaWire: Yes, I didn't want to have the same name as a Core Data entity.
`FeatureType` ?

Sources/Helpers/FeatureController.swift
KaterinaWire: Done.
I added this enum to DM and changed the name to `FeatureStatus` because it will be public and we might have other statuses.
Feature class in the DM now has a status property with this type.

Sources/Helpers/FeatureController.swift
KaterinaWire: Changed the return value type.

We can make all the next features configurable, but without any configuration.
E.g.
```
public enum DigitalSignature: Configurable {
        public static var name: String = "digital-signature"
        public struct Config: Codable {
        }
    }
```

Sources/Helpers/FeatureController.swift
KaterinaWire: Done

Sources/Helpers/FeatureController.swift
KaterinaWire: Fixed

Sources/Helpers/FeatureController.swift
KaterinaWire: Yes. 
Fixed

Sources/Helpers/FeatureController.swift
KaterinaWire: Added comment

Sources/Helpers/FeatureController.swift
KaterinaWire: Added logs

Sources/Helpers/FeatureController.swift
KaterinaWire: Removed

Tests/Helpers/FeatureControllerTest.swift
KaterinaWire: Fixed

Tests/Helpers/FeatureControllerTest.swift
KaterinaWire: Done

Sources/Helpers/FeatureController.swift
johnxnguyen: Just an idea: another way would be to split `Configurable` protocol into two:

```swift
protocol Configurable {
  typealias Config: Codable
}

protocol Named {
  static var name: String { get }
}

func status<T: Named>(for feature: T.Type) -> FeatureStatus

File: Sources/Helpers/FeatureController.swift [hunk 12]
```diff
@@ -0,0 +1,128 @@
...
+                                                        config: configuration.configData,
+                                                        context: moc)
+            
+            // TODO: Katerina make it more general for all features
+            if let appLockFeature = feature  {
+                NotificationCenter.default.post(name: FeatureController.needsToUpdateFeatureNotificationName, object: nil, userInfo: ["appLock" : appLockFeature])
+            }
+            
+        } catch {
+            zmLog.error("Failed to decode response: \(error)"); return
```
johnxnguyen, line 108: `return` is redundant.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,128 @@
...
+            let appLockFeature = Feature.createOrUpdate(appLock.name,
+                                                        status: appLock.schema.status,
+                                                        config: appLock.schema.configData,
+                                                        context: moc)
+            
+            if let appLockFeature = appLockFeature  {
+                NotificationCenter.default.post(name: FeatureController.needsToUpdateFeatureNotificationName, object: nil, userInfo: ["appLock" : appLockFeature])
+            }
+        } catch {
+            zmLog.error("Failed to decode response: \(error)"); return
```
johnxnguyen, line 125: `return` is redundant.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,128 @@
...
+
+// MARK: - Save to Core Data
+extension FeatureController {
+    public func save<T: Configurable>(_ feature: T.Type, data: Data) {
+        do {
+            let configuration = try JSONDecoder().decode(FeatureConfigResponse<T>.self, from: data)
+            let feature = Feature.createOrUpdate(feature.name,
+                                                        status: configuration.status,
+                                                        config: configuration.configData,
+                                                        context: moc)
```
johnxnguyen, line 100: Needs formatting.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,128 @@
...
+        guard let configData = Feature.fetch(T.name, context: moc)?.config else {
+                return nil
+        }
+        return try? JSONDecoder().decode(T.Config.self, from: configData)
+    }
+}
+
+// MARK: - Save to Core Data
+extension FeatureController {
+    public func save<T: Configurable>(_ feature: T.Type, data: Data) {
```
johnxnguyen, line 94: Is there a reason why we decode `data` here and not in the request strategy? To me I would expect the request strategy to handle decoding of response models.

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -0,0 +1,106 @@
...
+        """
+        let data = json.data(using: .utf8)!
+        
+        // When
+        sut.saveAllFeatures(data)
+       
+        // Then
+        let fechedFeature = Feature.fetch("applock",
+                                          context: self.uiMOC)
+        XCTAssertNotNil(fechedFeature)
```
johnxnguyen, line 58: It's worth asserting the values of `fetchedFeature`.

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -0,0 +1,106 @@
...
+        
+        let data = json.data(using: .utf8)!
+        
+        // When
+        sut.save(FeatureModel.AppLock.self, data: data)
+        
+        // Then
+        let fechedFeature = Feature.fetch("applock",
+                                          context: self.uiMOC)
+        XCTAssertNotNil(fechedFeature)
```
johnxnguyen, line 81: Also here, assert the values of `fetchedFeature`.

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -0,0 +1,106 @@
...
+               }
+               """
+        
+        let data = json.data(using: .utf8)!
+        
+        // When
+        sut.save(FeatureModel.AppLock.self, data: data)
+        
+        // Then
+        let featureStatus = sut.status(for: FeatureModel.AppLock.self)
```
johnxnguyen, line 102: If the method you want to test is `status(for:)`, then it should be under "when".

johnxnguyen (changes requested):

Tests/Helpers/FeatureControllerTest.swift
KaterinaWire: Fixed

Tests/Helpers/FeatureControllerTest.swift
KaterinaWire: Added

Tests/Helpers/FeatureControllerTest.swift
KaterinaWire: Added

Sources/Helpers/FeatureController.swift
KaterinaWire: Fixed

Sources/Helpers/FeatureController.swift
KaterinaWire: Removed

Sources/Helpers/FeatureController.swift
KaterinaWire: Removed

Sources/Helpers/FeatureController.swift
KaterinaWire: Yes, good idea.
But `Configurable` protocol should conform the `Named` protocol. When we save a feature, we should have access to the name and config.

Sources/Helpers/FeatureController.swift
KaterinaWire: To be honest I don't have a strong opinion about that. 
But if we decide to move it to the request strategy, that there is no need to keep "saving logic" inside the `FeatureController`. Currently we have access to `FeatureController` only from the request strategy.

Sources/Helpers/FeatureController.swift
johnxnguyen: You can use protocol composition to combine them, e.g `Named & Configurable`.

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -97,10 +101,9 @@ class FeatureControllerTest: MessagingTestBase {
         
         // When
         sut.save(FeatureModel.AppLock.self, data: data)
+        let featureStatus = sut.status(for: FeatureModel.AppLock.self)
```
johnxnguyen, line 104: The test is that is fetches the status. So `sut.save` should be part of the "given" section. This clearly indicates exactly which methods are being tested.

Sources/Helpers/FeatureController.swift
KaterinaWire: Changed to `Named & Configurable`

Sources/Helpers/FeatureController.swift
KaterinaWire: Moved data decoding to request strategy

Tests/Helpers/FeatureControllerTest.swift
KaterinaWire: Done

Sources/Helpers/FeatureController.swift
```diff
@@ -77,14 +75,14 @@ public class FeatureController {
         moc = managedObjectContext
     }
     
-    public func status<T: Named>(for feature: T.Type) -> Feature.Status {
-        guard let feature = Feature.fetch(T.name, context: moc) else {
+    public static func status<T: Named>(for feature: T.Type, context: NSManagedObjectContext) -> Feature.Status {
+        guard let feature = Feature.fetch(T.name, context: context) else {
```
johnxnguyen, line 79: Don't forget to switch the contexts queue, otherwise you'll get a crash. Something like:

```swift
context.performBlockAndWait {
  ...
}
```

Sources/Helpers/FeatureController.swift
```diff
@@ -77,14 +75,14 @@ public class FeatureController {
...
             return .disabled
         }
         return feature.status
     }
     
-    public func configuration<T: Configurable>(for feature: T.Type) -> T.Config? {
+    public func configuration<T: Configurable & Named>(for feature: T.Type) -> T.Config? {
```
johnxnguyen, line 85: Would this also need to be static?

Sources/Helpers/FeatureController.swift
```diff
@@ -94,34 +92,23 @@ public class FeatureController {
...
-                                                 config: configuration.configData,
-                                                 context: moc)
-            
-            // TODO: Katerina make it more general for all features
-            NotificationCenter.default.post(name: FeatureController.needsToUpdateFeatureNotificationName, object: nil, userInfo: ["appLock" : feature])
-            
-        } catch {
-            zmLog.error("Failed to decode response: \(error)")
-        }
+    public func save<T: Configurable & Named>(_ feature: T.Type, configuration: FeatureConfigResponse<T>) {
```
johnxnguyen, line 95: Does this need to be public? If only the request strategy will call it then better to keep it internal.

Sources/Helpers/FeatureController.swift
```diff
@@ -94,34 +92,23 @@ public class FeatureController {
...
-            let appLockFeature = Feature.createOrUpdate(appLock.name,
-                                                        status: appLock.schema.status,
-                                                        config: appLock.schema.configData,
-                                                        context: moc)
-            
-            NotificationCenter.default.post(name: FeatureController.needsToUpdateFeatureNotificationName, object: nil, userInfo: ["appLock" : appLockFeature])
-        } catch {
-            zmLog.error("Failed to decode response: \(error)")
-        }
+    public func saveAllFeatures(_ configurations: AllFeatureConfigsResponse) {
```
johnxnguyen, line 105: Does this need to be public? If only the request strategy will call it then better to keep it internal.

KaterinaWire: No, there is no need to make it public. 
Changed it

Sources/Helpers/FeatureController.swift
KaterinaWire: Changed it to internal

Sources/Helpers/FeatureController.swift
KaterinaWire: Yes.
Changed

Sources/Helpers/FeatureController.swift
KaterinaWire: Why will be crash?

Sources/Helpers/FeatureController.swift
marcomaddy1982: I guess the right name it would be `Feature` but to don't create confusion with the DM another option would be `FeatureObject`. It is just a suggestion

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,114 @@
...
+
+public enum FeatureModel {
+    public enum AppLock: Configurable, Named {
+        public static var name: String = "applock"
+        public struct Config: Codable {
+            let enforceAppLock: Bool
+            let inactivityTimeoutSecs: UInt
+            
+            private enum CodingKeys: String, CodingKey {
+                case enforceAppLock = "enforceAppLock"
```
marcomaddy1982, line 37: For the CodingKeys when the rawValue is equal to the key it is not necessary to specify it there. This could become 

```swift
private enum CodingKeys: String, CodingKey {
           case enforceAppLock
           case inactivityTimeoutSecs
 }
```

marcomaddy1982: Basically for the `FeatureConfigResponse` is done as I meant in the previous comment.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,114 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
```
marcomaddy1982, line 19: General thought: I would add more pragma mark to separate the code in this class.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,114 @@
...
+
+// MARK: - Save to Core Data
+extension FeatureController {
+    internal func save<T: Configurable & Named>(_ feature: T.Type, configuration: FeatureConfigResponse<T>) {
+        let feature = Feature.createOrUpdate(feature.name,
+                                             status: configuration.status,
+                                             config: configuration.configData,
+                                             context: moc)
+        
+        // TODO: Katerina make it more general for all features
```
marcomaddy1982, line 101: Thanks Katerina :-D. 💯

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+extension FeatureConfigResponse {
+    var configData: Data? {
+        return try? JSONEncoder().encode(config)
+    }
+}
+
+public struct AllFeatureConfigsResponse: Decodable {
```
marcomaddy1982, line 60: Why don't call it just `FeatureConfigsResponse` and remove the prefix `all`. But if @johnxnguyen approves this naming means that it is better.

Sources/Helpers/FeatureController.swift
```diff
@@ -0,0 +1,114 @@
...
+                                             config: configuration.configData,
+                                             context: moc)
+        
+        // TODO: Katerina make it more general for all features
+        NotificationCenter.default.post(name: FeatureController.needsToUpdateFeatureNotificationName, object: nil, userInfo: ["appLock" : feature])
+    }
+    
+    internal func saveAllFeatures(_ configurations: AllFeatureConfigsResponse) {
+        let appLock = (name: FeatureModel.AppLock.name, schema: configurations.applock)
+        let appLockFeature = Feature.createOrUpdate(appLock.name,
```
marcomaddy1982, line 107: Here you can use the method you created on top `func save<T: Configurable & Named>`  so you can remove the duplicated `Feature.createOrUpdate(...`

KaterinaWire: Yes, I could, but I need to post a notification with a `appLockFeature` inside the `userInfo`.
Currently `func save<T: Configurable & Named>`doesn't return the value.

Sources/Helpers/FeatureController.swift
KaterinaWire: Yes, good point.

Sources/Helpers/FeatureController.swift
johnxnguyen: Well, maybe we could simply move the contents of `FeatureModel` into an extension for `Feature`. The only purpose of the `FeatureModel` enum is to provide a namespace for the phantom types. `Feature` is a suitable name and the data model class is closely related to the phantom types so I don't see a problem with using it as a namespace. Thoughts?

Sources/Helpers/FeatureController.swift
johnxnguyen: Every managed object context is associated with a thread. The ui context runs on the main thread, and all other contexts run on their own private background thread. If you work with a context on a different thread Core Data will crash your app. It's part of the concurrency model for Core Data. You can read more about it [here](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext)

Sources/Helpers/FeatureController.swift
johnxnguyen: Adding to this, I would even go as far as to separate it into separate files 😄

Sources/Helpers/FeatureController.swift
johnxnguyen: I suggested to add the "all" prefix to make it easier to spot the difference:

```swift
FeatureConfigResponse
FeatureConfigsResponse
```

vs

```swift
FeatureConfigResponse
AllFeatureConfigsResponse
```

Sources/Helpers/FeatureController.swift
johnxnguyen: It is also unnecessary to define `CodingKeys` if the cases match the property names, the compiler will synthesize this for you.

Sources/Helpers/FeatureController.swift
marcomaddy1982: Ok, mine was just a thought :-)

Sources/Helpers/FeatureController.swift
marcomaddy1982: Make sense!!!

Sources/Helpers/FeatureController.swift
KaterinaWire: Removed CodingKeys

Sources/Helpers/FeatureController.swift
KaterinaWire: Done

Sources/Helpers/FeatureController.swift
KaterinaWire: Good idea. 
Moved `AppLock enum` to the  `Feature` extension.

Sources/Helpers/FeatureController.swift
KaterinaWire: Make sense.

marcomaddy1982 (approved):

Sources/Helpers/FeatureConfiguration/Configurations.swift
```diff
@@ -0,0 +1,51 @@
...
+        public static var name: String = "applock"
+        public struct Config: Codable {
+            let enforceAppLock: Bool
+            let inactivityTimeoutSecs: UInt
+        }
+    }
+}
+
+// MARK: - Feature Responses
+public struct FeatureConfigResponse<T: Configurable>: Decodable {
```
johnxnguyen, line 40: Does this need to be public?

Sources/Helpers/FeatureConfiguration/Configurations.swift
```diff
@@ -0,0 +1,51 @@
...
+    var status: Feature.Status
+    var config: T.Config?
+    
+    var configData: Data? {
+        return try? JSONEncoder().encode(config)
+    }
+}
+
+public struct AllFeatureConfigsResponse: Decodable {
```
johnxnguyen, line 49: Does this need to be public?

Sources/Helpers/FeatureConfiguration/FeatureController.swift
```diff
@@ -0,0 +1,71 @@
...
+
+public class FeatureController {
+    
+    public static let needsToUpdateFeatureNotificationName = Notification.Name("needsToUpdateFeatureConfiguration")
+
+    private(set) var moc: NSManagedObjectContext
+    
+    public init(managedObjectContext: NSManagedObjectContext) {
+        moc = managedObjectContext
+    }
```
johnxnguyen, line 29: Does this need to be public?

Sources/Helpers/FeatureConfiguration/FeatureController.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class FeatureController {
+    
+    public static let needsToUpdateFeatureNotificationName = Notification.Name("needsToUpdateFeatureConfiguration")
```
johnxnguyen, line 23: I think a better name would be something like `FeatureConfigDidChange`.

Tests/Helpers/FeatureControllerTest.swift
```diff
@@ -0,0 +1,110 @@
...
+        let configuration = try! JSONDecoder().decode(FeatureConfigResponse<Feature.AppLock>.self, from: data)
+        sut.save(Feature.AppLock.self, configuration: configuration)
+        
+        // When
+        let featureStatus = FeatureController.status(for: Feature.AppLock.self, managedObjectContext: self.uiMOC)
+        
+        // Then
+        XCTAssertEqual(featureStatus, .enabled)
+    }
+}
```
johnxnguyen, line 110: Missing a test for fetching a config.

KaterinaWire: Added

Sources/Helpers/FeatureConfiguration/FeatureController.swift
KaterinaWire: Changed

Sources/Helpers/FeatureConfiguration/FeatureController.swift
KaterinaWire: No

Sources/Helpers/FeatureConfiguration/Configurations.swift
KaterinaWire: No

Sources/Helpers/FeatureConfiguration/Configurations.swift
KaterinaWire: No

johnxnguyen (approved):

PR #255, opened 2020-11-17
[Feature] Add FeatureConfigRequestStrategy

KaterinaWire:
### Issues
Add a `FeatureConfigRequestStrategy` that is responsible for:
- make requests to the backend
- process response from the backend
- process update events from backend

Needs releases with:

- https://github.com/wireapp/wire-ios-data-model/pull/1066
- https://github.com/wireapp/wire-ios-data-model/pull/1067
- https://github.com/wireapp/wire-ios-request-strategy/pull/254


johnxnguyen (commented):
One issue with the strategy is that it doesn't know when it shouldn't generate a request.

For example, if the self user is a team member, then whenever `nextRequestIfAllowed()` is called (and it can be called several times), it will always return a request.

We wouldn't have this if we had a queue or set of feature names that need requests.

```swift
enum PendingItem {
  case singleFeature(name: String)
  case allFeatures
}

private var pendingItems: [PendingItem]

func nextRequestIfAllowed() -> ZMTransportRequest? {
  guard let pendingItem = pendingItems.removeFirst else { return nil }

  switch pendingItem {
    case let .singleFeature(name):
      ...
    case .allFeatures:
      ...
  }
}
```

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,125 @@
...
+//
+
+import Foundation
+
+@objcMembers
+public final class FeatureConfigRequestStrategy: AbstractRequestStrategy {
+    
+    public static let needsToFetchFeatureConfigNotificationName = Notification.Name("needsToFetchFeatureConfiguration")
+    
+    private var observers: [Any] = []
```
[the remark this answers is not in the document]
KaterinaWire: Changed

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,125 @@
...
+    private var fetchSingleConfigSync: ZMSingleRequestSync?
+    private var fetchAllConfigsSync: ZMSingleRequestSync?
+    private var featureController: FeatureController?
```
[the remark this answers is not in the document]
KaterinaWire: We expect them to be non nil. I've made them as implicitly unwrapped. The same approach we have in other strategies.

KaterinaWire:
> One issue with the strategy is that it doesn't know when it shouldn't generate a request.
> 
> For example, if the self user is a team member, then whenever `nextRequestIfAllowed()` is called (and it can be called several times), it will always return a request.
> 
> We wouldn't have this if we had a queue or set of feature names that need requests.
> 
> ```swift
> enum PendingItem {
>   case singleFeature(name: String)
>   case allFeatures
> }
> 
> private var pendingItems: [PendingItem]
> 
> func nextRequestIfAllowed() -> ZMTransportRequest? {
>   guard let pendingItem = pendingItems.removeFirst else { return nil }
> 
>   switch pendingItem {
>     case let .singleFeature(name):
>       ...
>     case .allFeatures:
>       ...
>   }
> }
> ```

Implemented

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,125 @@
...
+        "enforce_app_lock": true,
+         "inactivity_timeout_secs": 30
+     }
+     */
+    
+    public func didReceive(_ response: ZMTransportResponse, forSingleRequest sync: ZMSingleRequestSync) {
+        guard let responseData = response.rawData,
+            (response.result == .permanentError || response.result == .success) else {
+             return
+        }
```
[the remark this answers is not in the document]
KaterinaWire: Added logs

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
```diff
@@ -0,0 +1,69 @@
...
+    override func tearDown() {
+        mockApplicationStatus = nil
+        sut = nil
+        super.tearDown()
+    }
+    
+    // MARK: Request generation
+    
+    func testThatItGeneratesARequestToFetchAllFeatureConfigurations() throws {
+//        self.syncMOC.performGroupedAndWait { moc in
```
marcomaddy1982, line 49: Remove the comments if not necessary anymore please. :-)

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,152 @@
...
+                                                       groupQueue: managedObjectContext)
+        self.observerToken = NotificationInContext.addObserver(
+            name: FeatureConfigRequestStrategy.needsToFetchFeatureConfigNotificationName,
+            context: self.managedObjectContext.notificationContext,
+            object: nil) { [weak self] in
+                self?.requestConfig(with: $0)
+        }
+    }
+    
+    public override func nextRequestIfAllowed() -> ZMTransportRequest? {
```
marcomaddy1982, line 63: // MARK: - overrides

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,152 @@
...
+            return nil
+        }
+        return ZMTransportRequest(getFromPath: "/teams/\(teamId)/features/\(feature)")
+    }
+}
+
+// MARK: - Private methods
+extension FeatureConfigRequestStrategy {
+    private func requestConfig(with note: NotificationInContext) {
+        if let featureName = note.object as? String {
```
marcomaddy1982, line 144: maybe a `ternary operator` here. :-) I'm big when they could be used but not all the developers are fun of them. So feel free to choose your favourite style

KaterinaWire: I also like the ternary operator, but in this case I should trigger `RequestAvailableNotification.notifyNewRequestsAvailable(self)` after `if...else` statement.
With the ternary operator, I have to call `.notifyNewRequestsAvailable(self)` twice.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
```diff
@@ -0,0 +1,64 @@
...
+            
+            // then
+            XCTAssertEqual(request.path, "/teams/\(teamId)/features")
+            XCTAssertEqual(request.method, .methodGET)
+        }
+    }
+    
+    func testThatItGeneratesARequestToFetchASingleFeatureConfiguration() {
+        
+    }
```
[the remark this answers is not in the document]
KaterinaWire: Yes.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
```diff
@@ -0,0 +1,64 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import WireRequestStrategy
+
+class FeatureConfigRequestStrategyTests: MessagingTestBase {
```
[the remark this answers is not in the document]
KaterinaWire: Added

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
KaterinaWire: Removed

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,155 @@
...
+            object: nil) { [weak self] in
+                self?.requestConfig(with: $0)
+        }
+    }
+    
+    // MARK: - Overrides
+    public override func nextRequestIfAllowed() -> ZMTransportRequest? {
+        guard let _ = pendingItems.first else {
+            return nil
+        }
```
johnxnguyen, line 67: Suggestion:

```swift
guard !pendingItems.isEmpty else { return nil }
```

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,155 @@
...
+        let pendingItem = pendingItems.removeFirst()
+        switch pendingItem {
+        case let .singleFeature(featureName):
+            featureNames.append(featureName)
+            fetchSingleConfigSync.readyForNextRequestIfNotBusy()
+            return fetchSingleConfigSync.nextRequest()
+        case .allFeatures:
+            fetchAllConfigsSync.readyForNextRequestIfNotBusy()
+            return fetchAllConfigsSync.nextRequest()
+        }
```
johnxnguyen, line 78: I think you can move all of this into `request(for sync:)`, that way you won't need to have `featureNames`.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,155 @@
...
+        default:
+            return nil
+        }
+    }
+    
+    public func didReceive(_ response: ZMTransportResponse, forSingleRequest sync: ZMSingleRequestSync) {
+        guard response.result == .permanentError || response.result == .success else {
+            zmLog.debug("error downloading feature configuration (\(response.httpStatus))")
+            return
+        }
```
johnxnguyen, line 100: I don't understand why we still proceed if we have a permanent error.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
```diff
@@ -0,0 +1,115 @@
...
+    func testThatItDoesNotGenerateARequestWithoutPostingNotification() {
+        self.syncMOC.performGroupedAndWait { moc in
+            // when
+            let request = self.sut.nextRequestIfAllowed()
+            
+            // then
+            XCTAssertNil(request)
+        }
+    }
+}
```
johnxnguyen, line 115: This would me more accurate to say "test it does not generate a request if there are no pending items". Additionally, you should also check

- It consumes pending items (i.e if I have one pending item, then I generate one request, but not any more).
- It generates the correct request for the pending item

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -0,0 +1,155 @@
...
+        guard let responseData = response.rawData else {
+            return
+        }
```
johnxnguyen, line 103: This is also a kind of error, so we should log it too.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: Fixed

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: Yes, you are right. We shouldn't proceed it.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: Added logs

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
KaterinaWire: It means that pendingItems should be public, but it's kind of an inner implementation.
It `testThatItGeneratesARequestToFetchAllFeatureConfigurations` and `testThatItGeneratesARequestToFetchASingleFeatureConfiguration` I set different notifications and it checks the correct request.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: I don't think I can move this into `request(for sync:) `, because I have to figure out here `fetchSingleConfigSync.nextRequest()` or `fetchAllConfigsSync.nextRequest() ` should be  triggered.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
marcomaddy1982: My idea was to use the ternary operator just for the `if else` but just leave as it is. Don't worry about.

marcomaddy1982 (approved):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
johnxnguyen: One way you could do it (I think):

```swift
func nextRequest() -> ZMTransportRequest? {
  return fetchSingleConfigSync.nextRequest() ?? fetchAllConfigsSync.nextRequest()
}

func request(for sync: Sync) -> ZMTransportRequest? {
  guard let pendingItem = pendingItems.first else { return nil }

  switch (sync, pendingItem) {
  case (fetchSingleConfigSync, .singleFeature(let name)) :
    return ...
  case (fetchAllConfigsSyync, .allFeatures):
    return ...
  default:
    return nil
  }
}
```

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
johnxnguyen: Pending items doesn't have to be public, you can test internal things too. But in any case, you don't need to reference the pending items. To test that it consumes pending items, just test that you can generate one request but if you ask for the `nextRequest()` again it should return `nil`. The second point about "it generates the correct request..." is redundant, you're already testing this 👍

KaterinaWire: Done

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: Done

johnxnguyen (approved):

PR #256, opened 2020-11-23
[Feature] Changed the notification object

KaterinaWire:
## What's new in this PR?

- Made `FeatureConfigResponse` public because we can reuse this struct as userInfo in `FeatureController.featureConfigDidChange` notification. 
- Changed type of the `userInfo` in `FeatureController.featureConfigDidChange` notification.

### DIFF
--- a/Sources/Helpers/FeatureConfiguration/FeatureController.swift
+++ b/Sources/Helpers/FeatureConfiguration/FeatureController.swift
@@ -66,7 +66,7 @@ extension FeatureController {
             userInfo["configKey"] = config
         }
         
-        NotificationCenter.default.post(name: FeatureController.featureConfigDidChange, object: nil, userInfo: [Feature.AppLock.name : userInfo])
+        NotificationCenter.default.post(name: FeatureController.featureConfigDidChange, object: nil, userInfo: userInfo)
     }
     
     internal func saveAllFeatures(_ configurations: AllFeatureConfigsResponse) {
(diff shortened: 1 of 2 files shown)


johnxnguyen (changes requested):
I would advise against reusing the response model in this way, even if it contains exactly the information that we need to communicate. Its primary use is to model a network response. If we don't contain its use to this specific scope, then any changes that made to it in the future will cause breaking changes in possibly all projects from request strategy to the UI.

I would suggest instead to populate the user info with individual values, something like:

```swift
let userInfo = [
  nameKey: name
  statusKey: status,
  configKey: config
]
```

At least until we think of a better way to pass the config info around.

Sources/Helpers/FeatureConfiguration/FeatureController.swift
```diff
@@ -55,13 +55,18 @@ extension FeatureController {
...
-        if let featureConfig = feature.config {
-            config = try? JSONDecoder().decode(Feature.AppLock.Config.self, from: featureConfig)
+        if let featureConfig = feature.config,
+            let config = try? JSONDecoder().decode(Feature.AppLock.Config.self, from: featureConfig) {
+            userInfo["configKey"] = config
         }
-        let featureInfo = FeatureConfigResponse<Feature.AppLock>(status: feature.status, config: config)
-        NotificationCenter.default.post(name: FeatureController.featureConfigDidChange, object: nil, userInfo: [Feature.AppLock.name : featureInfo])
+        
+        NotificationCenter.default.post(name: FeatureController.featureConfigDidChange, object: nil, userInfo: [Feature.AppLock.name : userInfo])
```
johnxnguyen, line 69: You should pass in `userInfo` as the argument, not wrap it inside another dictionary.

johnxnguyen (changes requested):

johnxnguyen (approved):

PR #258, opened 2020-11-24
[Refactor] Feature configuration

johnxnguyen:
## What's new in this PR?

This PR contains updates for changes in https://github.com/wireapp/wire-ios-data-model/pull/1075

### Summary of changes

`FeatureController` methods to fetch `status` and `configuration` are no longer needed because features can now be accessed by an instance of `Team`:

```swift
let appLock = selfUser.team?.feature(for: Feature.AppLock.self)
appLock?.status
appLock?.config.isEnforced
```

---

`FeatureController` methods to save configs to Core Data have been replaced with `store<T: FeatureLike>(feature: T)`.

---

The response structures have been adjusted and moved to the `FeatureConfigRequestStrategy`.

---

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1075


Sources/Helpers/FeatureConfiguration/FeatureController.swift
```diff
@@ -27,45 +29,26 @@ public class FeatureController {
...
+        do {
+            try feature.store(for: team, in: moc)
+        }
+        catch {
+            zmLog.error("Failed to store feature config in Core Data: \(error.localizedDescription)")
+        }
+
         // TODO: Katerina make it more general for all features
-        NotificationCenter.default.post(name: FeatureController.featureConfigDidChange, object: nil, userInfo: [Feature.AppLock.name : feature])
+        NotificationCenter.default.post(
```
KaterinaWire, line 47: Should a notification be sent only if it was successfully saved?

johnxnguyen: Yes good point!

johnxnguyen: Fixed

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -35,16 +35,22 @@ public final class FeatureConfigRequestStrategy: AbstractRequestStrategy {
...
         super.init(withManagedObjectContext: managedObjectContext, applicationStatus: applicationStatus)
         
-        self.configuration = [.allowsRequestsWhileOnline,
-                              .allowsRequestsDuringQuickSync,
-                              .allowsRequestsWhileInBackground]
+        self.configuration = [
+            .allowsRequestsWhileOnline,
+            .allowsRequestsDuringQuickSync,
+            .allowsRequestsWhileInBackground
+        ]
```
KaterinaWire, line 53: I can't put a comment in the right place, but I thought you wanted to change 
`self.observerToken = NotificationInContext.addObserver(` (line #60) and not to use the `NotificationInContext`

johnxnguyen: I thought you wanted to keep it as `NotificationInContext`?

johnxnguyen: As discussed, we'll change the notification type, but I will address it in a separate PR.

KaterinaWire: Ok

KaterinaWire (approved):

marcomaddy1982:
@johnxnguyen this is more a curiosity than a comment, so no changes are requested. I noticed you used all around the `type constraint` `<T: FeatureLike>`, why don't just let all the object being of type `FeatureLike`. Is it just a syntax preference or there is some other advantage?

marcomaddy1982 (approved):

johnxnguyen:
> why don't just let all the object being of type `FeatureLike`

@marcomaddy1982 `FeatureLike` has an associated type which means you can only use it as a generic type constraint (the `Config`). The associated type is necessary because each feature declares its own `Config` type, otherwise indeed it would be easier to just pass around the existentials of `FeatureLike`.

Admittedly modeling these feature configs has been quite challenging because the `Feature` entity needs to store generic data (`configData` is of type `Data`), but we want use these instances with strong types.

marcomaddy1982:
> > why don't just let all the object being of type `FeatureLike`
> 
> @marcomaddy1982 `FeatureLike` has an associated type which means you can only use it as a generic type constraint (the `Config`). The associated type is necessary because each feature declares its own `Config` type, otherwise indeed it would be easier to just pass around the existentials of `FeatureLike`.
> 
> Admittedly modeling these feature configs has been quite challenging because the `Feature` entity needs to store generic data (`configData` is of type `Data`), but we want use these instances with strong types. The way we achieve this is to consider instances of `Feature` as simply containers for some feature, which we use to create a specific feature instance, such as `Feature.AppLock`.

:+1

PR #259, opened 2020-11-26
[Refactor] Use downstream object sync

johnxnguyen:
### Issues

The `FeatureConfigRequestStrategy` is not able identify which feature a single config request response is associated with. Currently it assumes all responses are for the app lock feature.

### Causes

There is not enough information passed to the callback method to process the response to be able to identify the feature. We could check inside the payload of the response to identify it, but in general a single request sync isn't not the most suitable type of sync for this task.

### Solutions

Use a `ZMDownstreamObjectSync`, which associates requests and responses to instances of `Feature`. All code related to managing a list of pending items can be removed, since the new sync is able to manage this itself.

Additionally, I provide a cleaner way to trigger fetching all configs that hides the notification object and instead favors a static method `Feature.triggerBackendRefreshForAllConfigs()`.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1077


Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -18,142 +18,168 @@
...
     // MARK: - Overrides
+
     public override func nextRequestIfAllowed() -> ZMTransportRequest? {
-        return fetchSingleConfigSync.nextRequest() ?? fetchAllConfigsSync.nextRequest()
+        return fetchAllConfigsSync.nextRequest() ?? fetchSingleConfigSync.nextRequest()
+    }
+
+    // MARK: - Methods
+
+    private func requestAllConfig() {
```
KaterinaWire, line 92: I didn't find where `requestAllConfig()` is used.

johnxnguyen: It's a leftover after I moved the logic to the setter of `needsToFetchAllConfigs`. It's deleted now.

KaterinaWire (approved):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -18,142 +18,161 @@
...
-                zmLog.error("Failed to decode feature config response: \(error)")
+        do {
+            let decoder = JSONDecoder()
+            let encoder = JSONEncoder()
+
+            switch feature.name {
+            case .appLock:
+                let config = try decoder.decode(ConfigResponse<Feature.AppLock>.self, from: responseData)
+                feature.status = config.status
+                feature.configData = try encoder.encode(config.config)
```
marcomaddy1982, line 124: I understood what you are doing here but probably `config.config` could be a bit confusing. Anyway feel free to leave as it is if you prefer.

marcomaddy1982 (approved):

PR #261, opened 2020-12-04
[Refactor] Replace GenericMessageNotificationRequestStrategy with ResetSessionRequestStrategy

typfel:
## What's new in this PR?

I'm introducing a dedicated strategy for the handling of sending session reset messages. This will allow me to more easily perform more operations when a session reset message successfully sends.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1083

## Notes

This PR is part of series of PRs which will introduce the ability to reset a session from a decryption error message.


Sources/Request Strategies/User Clients/ResetSessionRequestStrategy.swift
```diff
@@ -0,0 +1,77 @@
...
+            return
+        }
+        
+        genericMessageStrategy.schedule(message: GenericMessage(clientAction: .resetSession),
+                                        inConversation: converation)
+        { (response) in
+            
+            switch response.result {
+            case .success, .permanentError:
+                mutableUserClient[keyPath: keyPath] = false
```
typfel, line 70: `needsToNotifyOtherUserAboutSessionReset` is set to `false` here.

marcomaddy1982 (approved):

Sources/Object Syncs/Helpers/KeyPathObjectSync.swift
```diff
@@ -0,0 +1,94 @@
...
+            if object[keyPath: keyPath] {
+                if !pending.contains(object) {
+                    pending.insert(object)
+                    transcoder?.synchronize(object, keyPath: keyPath)
+                }
+            } else {
+                if pending.contains(object) {
+                    pending.remove(object)
+                }
+            }
```
David-Henner, line 76:
```suggestion
            } else if pending.contains(object) {
                 pending.remove(object)
            }
```

Sources/Request Strategies/User Clients/ResetSessionRequestStrategy.swift
typfel: Moved this logic inside the `KeyPathObjectSync` to enforce that it always gets done.

Sources/Object Syncs/Helpers/KeyPathObjectSync.swift
```diff
@@ -0,0 +1,96 @@
...
+        objects.forEach { object in
+            var mutableObject = object
+            
+            if object[keyPath: keyPath] {
+                if !pending.contains(object) {
+                    pending.insert(object)
+                    transcoder?.synchronize(object) {
+                        mutableObject[keyPath: self.keyPath] = false
+                    }
+                }
```
David-Henner, line 71: I'm thinking that these nested `if` statements could be combined into one as well

David-Henner (approved):
Approved with the suggestion below. feel free to merge without applying

Sources/Object Syncs/Helpers/KeyPathObjectSync.swift
typfel: Yes, I could but I don't think it's more readable since you need think to understand that it doesn't affect the `else if` statement.

marcomaddy1982 (approved):

PR #262, opened 2020-12-09
[Refactor] Set needsToNotifyUser in Feature

KaterinaWire:
## What's new in this PR?

After receiving a response from BE, set `needsToNotifyUser`  to `true` if we know that this feature has been a change (property `enforceAppLock`).

Needs releases with:

- [https://github.com/wireapp/wire-ios-data-model/pull/1085 ]


Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -119,9 +119,10 @@ extension FeatureConfigRequestStrategy: ZMDownstreamTranscoder {
 
             switch feature.name {
             case .appLock:
-                let config = try decoder.decode(ConfigResponse<Feature.AppLock>.self, from: responseData)
-                feature.status = config.status
-                feature.configData = try encoder.encode(config.config)
+                let response = try decoder.decode(ConfigResponse<Feature.AppLock>.self, from: responseData)
+                feature.needsToNotifyUser = mandatoryAppLockHasBeenChanged(lastConfigData: feature.configData, newConfig: response.config)
```
typfel, line 123: Wouldn't it be easier to update `needsToNotifyUser` in a `didSet` of `configData`?

typfel (changes requested):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: Yes, it would be easier

PR #269, opened 2021-01-06
[Fix] Don't force unwrap optional timestamp

typfel:
### Issues

The app would crash upon receiving a message which is marked as transient and can't be decrypted.

### Causes

We ignore the timestamp for transient update events so that property will return `nil`. When we try to add the "Decryption Failed" system message we force unwrap the timestamp and crash.

### Solutions

Don't force unwrap the timestamp since `appendDecryptionFailedSystemMessage()` accepts an optional timestamp.


KaterinaWire (approved):

marcomaddy1982 (approved):

typfel:
@zenkins release patch

PR #287, opened 2021-02-25
[Fix] Some strategies not running while waiting for the web socket

typfel:
### Issues

The client fetching strategy wouldn't run after the changes made in https://github.com/wireapp/wire-ios-sync-engine/pull/1408

### Causes

We now correctly report the sync state to `.waitingForWebsocket` but the strategies weren't configured to run in that state.

### Solutions

Allow all strategies to run in `.waitingForWebsocket` if they are already allowed to run during the quick sync.


johnxnguyen (approved):

David-Henner (approved):

typfel:
@zenkins release patch

PR #290, opened 2021-03-04
[Fix] Improve sending in large groups

typfel:
### Issues

Sending the initial message in large groups is so slow that the message expires before it finish sending.

### Causes

The first time you send message to another user/client we first fetch the clients metadata. If you send your first message in group with several hundred members that could easily result in thousands in requests to fetch client metadata. The max group limit is 500 and every user could potentially have 8 clients which means we could potentially be performing 4000 requests.

`GET /users/<uid>/clients/<uid>`

Completing these requests takes a while, and the 30 second message expiration limit will most likely be reached before that happens.

### Solutions

Fetch the client metadata in batches using the new `POST /users/list-clients` endpoint. Since everyone might not run the latest version of the backend we fallback to the old way fetching client metadata if the endpoint doesn't exist.


Sources/Request Strategies/User Clients/FetchingClientRequestStrategyTests.swift
```diff
@@ -118,24 +119,174 @@ extension FetchClientRequestStrategyTests {
...
+
+            // THEN
+            XCTAssertEqual(self.sut.nextRequest()?.path, "/users/list-clients")
+        }
+    }
+
+    func testThatItUpdatesTheClient_WhenReceivingTheBatchResponse() {
+        var client: UserClient!
+        syncMOC.performGroupedBlockAndWait {
+            // GIVE
```
johnxnguyen, line 145:
```suggestion
            // GIVEN
```

johnxnguyen (approved):

marcomaddy1982 (approved):

typfel:
@johnxnguyen @marcomaddy1982 I had to make one more change https://github.com/wireapp/wire-ios-request-strategy/pull/290/commits/ad00814ba92f27dc0d1ad327b0362058ecf84408

I realised that fallback switch on 404 was only working by accident since they were getting re-scheduled due to other unrelated changes. To make this robust and work in all scenarios I need to explicitly re-schedule all the clients we failed to fetch when receiving the 404.

typfel:
@zenkins release minor

johnxnguyen (approved):

PR #291, opened 2021-03-15
[Fix] deviceClass was incorrectly marked as non-optional

typfel:
### Issues

Message can't be sent in a conversation which contains a client which doesn't have its deviceClass defined.

### Causes

`deviceClass` was annotated as non-optional which causes us to reject the whole response when fetching client metadata which contains a client which doesn't have the device class set.

### Solutions

Mark the `deviceClass` as optional. This will allow us to process all the clients with an assigned device class and do nothing when it's missing.

### DIFF
--- a/Sources/Payloads/Payload+Processing.swift
+++ b/Sources/Payloads/Payload+Processing.swift
@@ -22,7 +22,10 @@ extension Payload.UserClient {
     func update(_ client: WireDataModel.UserClient) {
         client.needsToBeUpdatedFromBackend = false
 
-        guard client.user?.isSelfUser == false else { return }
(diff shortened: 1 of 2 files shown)


KaterinaWire:
@typfel Could you change the `develop`(base) branch to the release branch?

typfel:
> @typfel Could you change the `develop`(base) branch to the release branch?

This is framework release, I'll make another PR for bumping the components.

KaterinaWire (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #292, opened 2021-03-17
[Fix] Update the conversation security level in the updateEvents decryption method

KaterinaWire:
### Issues
When a new client posts a message in a verified conversation, there is no indication that this is a new client.

### Solutions
Since we have information about the new client when decrypting events, we can also update the security level at this point.(add `updateSecurityLevelAfterDiscovering(...)` method to the event decryption method).

## Notes
https://wearezeta.atlassian.net/browse/SQCORE-539

### DIFF
--- a/Sources/Helpers/EncryptionSessionDirectory+UpdateEvents.swift
+++ b/Sources/Helpers/EncryptionSessionDirectory+UpdateEvents.swift
@@ -71,6 +71,7 @@ extension EncryptionSessionsDirectory {
         if createdNewSession {
             selfUser.selfClient()?.decrementNumberOfRemainingKeys()
             selfUser.selfClient()?.addNewClientToIgnored(senderClient)
+            selfUser.selfClient()?.updateSecurityLevelAfterDiscovering(Set(arrayLiteral: senderClient))
         }
         
         return decryptedEvent


typfel:
I think we should add a test which asserts that we call `updateSecurityLevelAfterDiscovering` when we discover a new client, but I see that the existing tests for `EncryptionSessionsDirectory` are quite bad so perhaps it's better break that out into a separate ticket so that we can move on with the RC.

typfel (approved):

marcoconti83 (commented):
The PR makes sense but did we figure out why this broke in the first place? It used to work, didn't it? If yes, I think it's important to find out why it broke before we apply this fix

typfel:
> The PR makes sense but did we figure out why this broke in the first place? It used to work, didn't it? If yes, I think it's important to find out why it broke before we apply this fix

Yes, it broke after the switch to the batch fetching client metadata endpoint. The old single fetch strategy always called `updateSecurityLevelAfterDiscovering` when processing the response while the new one only does it when discovering a new client. 

Previously:
1. Receive event and add client and create session
2. Fetch client metadata `GET /users/<uid>/clients/<cid>` because `needsUpdateFromBackend` is `true`
3. Receive response and call `updateSecurityLevelAfterDiscovering`

After changes:
1. Receive event and add client and create session
2. Fetch client metadata `POST /users/list-clients` because `needsUpdateFromBackend` is `true`
3. Receive response but we don't call `updateSecurityLevelAfterDiscovering` because the client already existed and we already have a session with it.

marcoconti83:
It has been a long time so the details are fuzzy. But if calling `updateSecurityLevelAfterDiscovering(...)` has no side effect if the client was already discovered and the level updated (as in, it will result in the same level if called multiple times), then I see no downside with this approach. 👍

marcoconti83 (approved):

typfel:
> updateSecurityLevelAfterDiscovering

No, nothing happens if the security level is already updated.

KaterinaWire:
@zenkins release patch

PR #305, opened 2021-05-28
[Refactor] Simplify MissingClientsRequestStrategy

typfel:
## What's new in this PR?

Simplify `MissingClientsRequestStrategy` and make it using type safe structs for decoding and encoding payloads. This is done  in preparation to support the federation aware `/users/list-prekeys` in addition to the existing legacy `/users/prekeys` endpoint.

PR #306, opened 2021-05-31
[Feature] Support new federation aware endpoint for fetching prekeys

typfel:
## What's new in this PR?

Support new federation aware endpoint for fetching prekeys.

- If the server doesn't yet support the new endpoint we'll fallback to the old legacy endpoint.
- The pagination of prekey fetching has been simplified by limiting the number of clients we fetching instead of the number users.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/305

## Notes

TODO: I'll add a way fetch to fetch users and clients locally using a qualified ID (ID and domain) .


Sources/Payloads/Payload+Processing.swift
```diff
@@ -91,6 +91,37 @@ extension Payload.PrekeyByUserID {
 
 }
 
+extension Payload.PrekeyByQualifiedUserID {
+
+    func establishSessions(with selfClient: UserClient, context: NSManagedObjectContext) -> Bool {
+        for (domain, prekeyByUserID) in self {
```
johnxnguyen, line 97: This method duplicates a lot of code of a method with the same name found in the extension of `Payload.PrekeyByUserID` above. It looks like only this outer loop is different. So you could probably do something like:

```swift
func establishSessions(with selfClient: UserClient, context: NSManagedContext) -> Bool {
  for _, prekeyByUserID in self {
    prekeyByUserID.establishSessions(with: selfClient, context: context)
  )

  return ...
}
```

Sources/Request Strategies/User Clients/MissingClientsRequestFactory.swift
```diff
@@ -38,6 +38,25 @@ extension Set where Set.Element == UserClient {
...
+            else {
+                return
+            }
+
+            var users = result[domain] ?? Payload.ClientListByUserID()
+            var clientList = users[userID] ?? []
+            clientList.append(clientID)
+            users[userID] = clientList
+            result[domain] = users
+        }
```
johnxnguyen, line 58: Dictionary defaults may be more readable here:

```swift
var users = result[domain, default: Payload.ClientListByUserID()]
users[usersID, default: []].append(clientID)
result[domain] = users
```

Sources/Request Strategies/User Clients/MissingClientsRequestStrategyTests.swift
```diff
@@ -190,7 +239,83 @@ class MissingClientsRequestStrategyTests: MessagingTestBase {
...
+                          "Unrecognized user")
+            XCTAssertTrue([user.remoteIdentifier!.transportString(),
+                           self.otherUser.remoteIdentifier!.transportString()].contains(secondEntry.key),
+                          "Unrecognized user")
+            XCTAssertNotEqual(firstEntry.key, secondEntry.key)
+            let expectedClients1 = [self.otherClient.remoteIdentifier!]
+            let expectedClients2 = [otherClient2.remoteIdentifier!]
+            XCTAssertTrue(firstEntry.value == expectedClients1 || firstEntry.value == expectedClients2, "Not matching clients")
+            XCTAssertTrue(secondEntry.value == expectedClients1 || secondEntry.value == expectedClients2, "Not matching clients")
+            XCTAssertNotEqual(firstEntry.value, secondEntry.value)
```
johnxnguyen, line 262: Perhaps this could be made more readable by putting values in a set and comparing the sets:

```swift
XCTAssertEqual(Set(firstEntry.value, secondEntry.value), Set(expectedClients1, expectedClients2))
```

johnxnguyen (approved):

KaterinaWire (approved):

PR #308, opened 2021-06-03
[Feature] Support federation aware API for fetching user profiles 

typfel:
## What's new in this PR?

Support federation aware API for fetching user profiles with fallback to the legacy API if the new endpoint is not available. I'm implementing a new request strategy for this endpoint since the existing strategy [ZMUserTranscoder](https://github.com/wireapp/wire-ios-sync-engine/blob/3e295f800c7819a312b9fe5840edec19898232b8/Source/Synchronization/Transcoders/ZMUserTranscoder.m) is written in objective-C and currently lives in SE.

This will be a multi-part PR (see TODO list).

## Dependencies

Needs releases with:

- [ ] N/A

## Todos

- [x] Introduce protocol for taking part of the slow sync
- [x] Add federation aware user profile strategy with tests


Sources/Protocols/SyncProgress.swift
```diff
@@ -0,0 +1,93 @@
...
+    public var isSyncing : Bool {
+        return self != .done
+    }
+
+    public var nextPhase: SyncPhase {
+        return SyncPhase(rawValue: rawValue + 1) ?? .done
+    }
+
+    public static var lastSlowSyncPhase: SyncPhase {
+        return .fetchingLabels
```
David-Henner, line 50: isn't `fetchingMissedEvents` the last phase? Or is it just not part of the slow sync?

David-Henner (approved):

Sources/Protocols/SyncProgress.swift
typfel: `fetchingMissedEvents` is what we call the quick sync so it's not part of the slow sync.

PR #310, opened 2021-06-21
[Feature] Fetch users with qualified ID when available

typfel:
## What's new in this PR?

- Update to use the new API for fetching and creating users introduced in https://github.com/wireapp/wire-ios-data-model/pull/1168
- In contexts where we have access to a qualifiedID include the domain when fetching a user.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1168

## Notes

Not all endpoints and events have been updated to include qualified IDs so these will implemented in a future PR.


johnxnguyen (approved):

Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
```diff
@@ -242,13 +245,13 @@ fileprivate final class UserClientByQualifiedUserIDTranscoder: IdentifierObjectS
...
+        for (domain, users) in payload {
             for (userID, clientPayloads) in users {
                 guard
                     let userID = UUID(uuidString: userID),
-                    let user = ZMUser.fetchAndMerge(with: userID,
-                                                    createIfNeeded: true,
-                                                    in: managedObjectContext)
+                    let user = ZMUser.fetch(with: userID,
+                                            domain: domain,
+                                            in: managedObjectContext)
```
David-Henner, line 254: should it be `fetchOrCreate` here since previous API used `createIfNeeded: true` ?

typfel: Yes, good catch!

David-Henner (approved):

PR #316, opened 2021-06-28
[Feature] Integrate new endpoint for sending qualified messages

typfel:
## What's new in this PR?

Integrate the new endpoint for sending qualified messages in the `ClientMessageTranscoder`. This was a proof-concept implementation to test the new endpoint. 

There are several other request strategies which will also needs to be updated to support the qualified endpoint. Doing this in the same way it was done for the `ClientMessageTranscoder ` would result in large amounts of code duplication. I therefore decided to refactor the message sending code a bit so that he logic can be shared. This will happen in a follow up PR.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1171
- [x] https://github.com/wireapp/wire-ios-protos/pull/47


Sources/Request Strategies/Client Message/ClientMessageTranscoderTests.swift
```diff
@@ -34,6 +34,7 @@ class ClientMessageTranscoderTests: MessagingTestBase {
         mockAttachmentsDetector = MockAttachmentDetector()
         LinkAttachmentDetectorHelper.setTest_debug_linkAttachmentDetector(mockAttachmentsDetector)
         sut = ClientMessageTranscoder(in: syncMOC, localNotificationDispatcher: localNotificationDispatcher, applicationStatus: mockApplicationStatus)
+        sut.isFederationEndpointAvailable = false
```
KaterinaWire, line 37: Do we have tests in `ClientMessageTranscoderTests` for `sut.isFederationEndpointAvailable = true`?

typfel: No, because I'll introduce a `ProteusMessageSync` in a follow up PR which will have these tests. This will be the only code path which generates the requests for sending messages, which means I only need to write these tests once.

KaterinaWire (approved):

David-Henner (approved):

PR #317, opened 2021-06-29
[Fix] Notify that user cannot send a message because one of the devices doesn't support LH

KaterinaWire:
### Issues
When a user who is under LH has one of the devices that doesn't support LH (not updated client), there's an 412 error from BE and the user is unable to send a message.

### Causes
All clients are not updated.

### Solutions
Handle the `403` error with `"missing-legalhold-consent"` label and notify the user.

## Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-526


Sources/Request Strategies/Client Message/ClientMessageTranscoder.swift
```diff
@@ -230,6 +230,16 @@ extension ClientMessageTranscoder {
...
         }
+
+        if response.httpStatus == 412,
+           let label = response.payloadLabel(), label == "missing-legalhold-consent" {
+            managedObjectContext.zm_userInterface.performGroupedBlock { [weak self] in
+                guard let `self` = self else { return }
+                NotificationInContext(name: ZMConversation.failedToSendMessageNotificationName,
+                                      context: self.managedObjectContext.notificationContext).post()
+            }
+        }
```
johnxnguyen, line 241: Suggestion to clean this up a little: 

```swift
if response.httpStatus == 412 && response.payloadLabel() == "missing-legalhold-consent" {
  managedObjectContext.zm_userInterface.performGroupedBlock { [weak self] in
    guard let context = self?.managedObjectContext.notificationContext else { return }
    NotificationInContext(name: ZMConversation.failedToSendMessageNotificationName, context: context).post()
  }
}
```

johnxnguyen (changes requested):
It would be good to add a test that the notification is sent.

typfel:
Is there any specs for how this should behave? I'm asking because the `ClientMessageTranscoder` is not the only place where you can receive this error.

ClientMessageTranscoder
AssetClientMessageRequestStrategy
LinkPreviewUploadRequestStrategy
GenericMessageRequestStrategy which is used in CallingRequestStrategy, AvailabilityRequestStrategy, DeliveryReceiptRequestStrategy

All send encrypted messages which can then in theory receive this error.

I guess mostly when sense show the alert when the user initiates the message sending and not when a delivery receipt is sent but that's up for discussion I guess.

KaterinaWire:
> LinkPreviewUploadRequestStrategy

We have only this ticket https://wearezeta.atlassian.net/browse/SQSERVICES-524. 
I agree that it should cover sending not only a text message, but assets as well.

About the `LinkPreviewUploadRequestStrategy`, we won't trigger it as the message will not be sent. Correct?
The calling is not a part of this task, but I will add this question to the parent ticket, because it's relevant to other platforms.

So, as I understand it, only 2 strategies (ClientMessageTranscoder, AssetClientMessageRequestStrategy) are required to handle this error. @typfel  Are you agree?

Sources/Request Strategies/Client Message/ClientMessageTranscoder.swift
KaterinaWire: Done

johnxnguyen (approved):

typfel (approved):

KaterinaWire:
@zenkins release minor

PR #319, opened 2021-07-05
[Refactor] Message sending

typfel:
## What's new in this PR?

Refactor the message sending so that all strategies rely on a single code path for sending messages. This reduces code and test duplication.

- Introduce `ProteusMessageSync` as the only way to encrypt and send proteus messages. This sync object is federation aware and will fall back to legacy endpoint if federation is not supported.
- Introduce new sync classes for syncing modified keys and newly inserted objects.
- Refactor `ClientMessageTranscoder`, `AssetClientMessageRequestStrategy`, `LinkPreviewUploadStrategy`, `DeliveryReceiptRequestStrategy` and `ResetSessionRequestStrategy` to use to `ProteusMessageSync`

After this changes I will delete:

- `OTREntityTranscoder`
- `GenericMessageRequestStrategy`

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/1173

## Notes

The last strategy which doesn't depend on `ProteusMessageSync` is `AvailabilityRequestStrategy`. It will be updated in a separate PR since it uses the broadcast endpoint which `ProteusMessageSync` doesn't yet support.


Sources/Object Syncs/Helpers/InsertedObjectSyncTests.swift
```diff
@@ -0,0 +1,132 @@
+//
+//  InsertedObjectSyncTests.swift
+//  WireRequestStrategyTests
+//
+//  Created by Jacob Persson on 29.06.21.
+//  Copyright © 2021 Wire GmbH. All rights reserved.
+//
```
David-Henner, line 7: Header should be changed to the Wire header

Sources/Request Strategies/Client Message/ClientMessageRequestStrategy.swift
```diff
@@ -0,0 +1,162 @@
...
+        let notDelivered = NSPredicate(format: "%K == FALSE", DeliveredKey)
+        let notExpired = NSPredicate(format: "%K == 0", ZMMessageIsExpiredKey)
+        let fromSelf = NSPredicate(format: "%K == %@", ZMMessageSenderKey, ZMUser.selfUser(in: context))
+        return NSCompoundPredicate(andPredicateWithSubpredicates: [notDelivered, notExpired, fromSelf])
+    }
+
+    public init(withManagedObjectContext managedObjectContext: NSManagedObjectContext,
+                         localNotificationDispatcher: PushMessageHandler,
+                         applicationStatus: ApplicationStatus) {
+
```
David-Henner, line 39: Commenting for indentation here

David-Henner (approved):

Sources/Object Syncs/Helpers/InsertedObjectSync.swift
```diff
@@ -0,0 +1,91 @@
...
+    ///                 on the backend.
+    ///
+    func insert(object: Object, completion: @escaping () -> Void)
+
+}
+
+/**
+ InsertedObjectSync synchronizes objects which has been inserted locally but does yet exist on the backend.
+
+ This only works for core data entities which inherit from `ZMManagedObject`. The Rule for when an object does
```
johnxnguyen, line 40:
```suggestion
 This only works for core data entities which inherit from `ZMManagedObject`. The rule for when an object does
```

Sources/Object Syncs/Helpers/InsertedObjectSync.swift
```diff
@@ -0,0 +1,91 @@
...
+ */
+class InsertedObjectSync<Transcoder: InsertedObjectSyncTranscoder>: NSObject, ZMContextChangeTracker {
+
+    let insertPredicate: NSPredicate
+    var pending: Set<Transcoder.Object> = Set()
+
+    weak var transcoder: Transcoder?
+
+    /// - Parameters:
+    ///   - insertPredicate: Predicate which determine when an object is only exist locally. If omitted
```
johnxnguyen, line 52:
```suggestion
    ///   - insertPredicate: Predicate which determine when an object only exists locally. If omitted
```

Sources/Object Syncs/Helpers/InsertedObjectSync.swift
```diff
@@ -0,0 +1,91 @@
...
+    ///                     `predicateForObjectsThatNeedToBeInsertedUpstream()` will be used.
+    init(insertPredicate: NSPredicate? = nil) {
+        self.insertPredicate = insertPredicate ?? Transcoder.Object.predicateForObjectsThatNeedToBeInsertedUpstream()!
+    }
+
+    func objectsDidChange(_ objects: Set<NSManagedObject>) {
+        let trackedObjects = objects.compactMap({ $0 as? Transcoder.Object})
+        let insertedObjects = trackedObjects.filter({ insertPredicate.evaluate(with: $0) })
+        let removedObjects = trackedObjects.filter({ !insertPredicate.evaluate(with: $0) })
```
johnxnguyen, line 61: `trackedObjects` may be large, so it might be expensive to filter the array twice, which would mean we evaluate the predicate for each item twice.

One way we could avoid this is to partition the array using the predicate:

```swift
var trackedObjects = objects.compactMap { $0 as? Transcoder.Object }
let indexOfSecondPartition = trackedObject.partition(by: insertPredicate.evaluate)
let insertedObjects = trackedObjects[indexOfSecondPartition...]
let removedObjects = trackedObjects[..<indexOfSecondPartition]
```

johnxnguyen (approved):

PR #320, opened 2021-07-06
[Chore] Revert support for federation aware API for fetching prekeys

typfel:
## What's new in this PR?

Revert support for the federation aware API for fetching prekey since it's incompatible with older server installations.


David-Henner (approved):

typfel:
@zenkins release minor

PR #331, opened 2021-07-23
[Refactor] Feature configs

johnxnguyen:
## What's new in this PR?

See https://github.com/wireapp/wire-ios-data-model/pull/1183 for a more thorough overview of the refactoring.

In this PR we...
- remove the redundant `FeatureController`. It wasn't really doing anything.
- Update the response structures to no longer rely on the non existent `FeatureLike` protocol.
- Update the request strategy for changes in data model
- Update tests.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1183


Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -164,7 +159,10 @@ extension FeatureConfigRequestStrategy: ZMSingleRequestTranscoder {
 
         do {
             let allConfigs = try JSONDecoder().decode(AllConfigsResponse.self, from: responseData)
-            featureController.store(feature: allConfigs.applock.asFeature, in: team)
+
+            let featureService = FeatureService(context: managedObjectContext)
+            featureService.storeAppLock(.init(configResponse: allConfigs.applock))
```
KaterinaWire, line 164: We are not currently using it, but it should be a generic `store()` method, not just `storeAppLock`.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -100,7 +96,7 @@ extension FeatureConfigRequestStrategy: ZMDownstreamTranscoder {
     }
 
     private func requestToFetchConfig(for feature: Feature) -> ZMTransportRequest? {
-        guard let teamId = feature.team?.remoteIdentifier?.transportString() else { return nil }
+        guard let teamId = team?.remoteIdentifier?.transportString() else { return nil }
```
KaterinaWire, line 99: This strategy still uses the endpoint with teamId. Are you going to introduce another PR with a change in the endpoint?
(However, `/feature-configs/appLock` doesn't work now)

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
johnxnguyen: I've opted to have separate store methods for each feature because it's much simpler. Having a generic method capable of storing any feature requires some kind of abstraction common to all features, which is what we tried to do with `FeatureLike` but admittedly it was overly complicated.

Do you have some concerns about this simpler approach?

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -174,23 +172,29 @@ extension FeatureConfigRequestStrategy: ZMSingleRequestTranscoder {
...
+private struct AllConfigsResponse: Decodable {
+
+    var applock: DynamicConfigResponse<Feature.AppLock.Config>
 
-    struct ConfigResponse<T: FeatureLike>: Decodable {
+}
 
-        let status: Feature.Status
-        let config: T.Config
+private struct ConfigResponse: Decodable {
```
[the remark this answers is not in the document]
johnxnguyen: Yes, and `DynamicConfigResponse` is for those with a decodable config.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
johnxnguyen: It's not strictly necessary now, but indeed we may want to do that as soon as we're sure it is working.

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: The response will be something like this:
```
{
 "appLock": {
        "status": "enabled"
 }
 "legalHold": {
        "status": "enabled",
        "config": {
              "isForced": true,
              "timeout": 555
          }
 }
 "fileSharing": {
        "status": "enabled"
 }
}
```
and we should store several features.

KaterinaWire: My concern is that for every new feature we have to create `storeFeatureName()` and `fetchFeatureName() `methods.

johnxnguyen: > right?

Yes, that is more or less how it would look.

> My concern is that for every new feature we have to create storeFeatureName() and fetchFeatureName() methods.

This is the trade off to have a simpler data structures for the features. The `FeatureLike` protocol doesn't really handle features that don't have their own `Config`. How should this be expressed? An optional `config` field? But this leads to a lot of unwrapping when you know there should be a config value. Or declaring `config: Never`? But `Never` doesn't conform to `Codable` and it's not really clear how it should. 

It felt like it was forcing an abstraction over was is essentially a heterogeneous set of possible features. Indeed they share some commonalities, but the possible presence of a completely new `Config` type made it a bit awkward. The cost of removing this abstraction is that we can't write polymorphic methods for these feature objects (like storing), but I think it's a small cost for the gained simplicity. That's my take on it.

KaterinaWire (approved):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
David-Henner: If you wanted to have only one method for `storeXXX()` and one for `fetchXXX()`, could you use an enum? I suppose it would look something like that 

```
featureService.store(.appLock(.init(configResponse: allConfigs.appLock)))
``` 

With this I think you could avoid duplicating some of the logic from `store` and `fetch` methods

David-Henner (approved):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
johnxnguyen: The problem, unfortunately, is that the return type is different for each feature.

johnxnguyen:
@zenkins release major

PR #343, opened 2021-08-15
[Feature] Enforce file sharing

KaterinaWire:
## What's new in this PR?

This is a feature branch containing all changes related to the file sharing feature flag. This feature flag should enable or disable for all team members the ability to send / receive files (files, images, gif) and record audio / video.

Needs releases with:

- [ wireapp/wire-ios-transport/pull/200 ] 
- [ wireapp/wire-ios-data-model/pull/1207 ]

## Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-613

### Attachments
https://zpl.io/beygezY


Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
```diff
@@ -169,6 +179,61 @@ extension FeatureConfigRequestStrategy: ZMSingleRequestTranscoder {
...
+    public var config: Data?
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let nestedContainer = try container.nestedContainer(keyedBy: ConfigKeys.self, forKey: .data)
+
+        name = try container.decode(Feature.Name.self, forKey: .name)
+        status = try nestedContainer.decode(Feature.Status.self, forKey: .status)
+        let configDict = try nestedContainer.decodeIfPresent([String: String].self, forKey: .config)
+        config = try? JSONEncoder().encode(configDict)
```
johnxnguyen, line 224: You could also decode it as a `String` then convert it to `Data`, it would save you from having to decode and re-encode.

johnxnguyen (approved):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategy.swift
KaterinaWire: I removed decode and re-encode, but changed this part for applock. There is also a new test.
```
switch name {
        case .appLock:
            config = try nestedContainer.decodeIfPresent(Feature.AppLock.Config.self, forKey: .config).payloadData()
        default:
            return
        }
```

David-Henner (approved):

KaterinaWire:
@zenkins release major

PR #349, opened 2021-08-23
[Feature] Support for disabling federation

typfel:
## What's new in this PR?

Introduce a `FederationAware` protocol which can be implemented by request strategies to declare that they interacting with federation endpoints and provide an option to disable their usage.

### DIFF
--- a/Sources/Protocols/SharedProtocols.swift
+++ b/Sources/Protocols/SharedProtocols.swift
@@ -20,13 +20,22 @@
 import Foundation
 
 
-@objc public protocol ClientRegistrationDelegate : NSObjectProtocol  {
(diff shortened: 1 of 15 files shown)


Sources/Protocols/SharedProtocols.swift
```diff
@@ -30,5 +30,14 @@ import Foundation
 
 }
 
+/// Request strategies which adopt this protocol declare that they can will talk to federation aware endpoints.
```
johnxnguyen, line 33:
```suggestion
/// Request strategies which adopt this protocol declare that they can talk to federation aware endpoints.
```

johnxnguyen (approved):

KaterinaWire (approved):

PR #350, opened 2021-08-24
[Feature] Add tests

KaterinaWire:
## What's new in this PR?

Add and fix tests with a new endpoint.

Needs releases with:

- [ wireapp/wire-ios-data-model/pull/1213 ]


johnxnguyen (approved):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
```diff
@@ -79,21 +77,27 @@ class FeatureConfigRequestStrategyTests: MessagingTestBase {
...
 
-            // Then
-            XCTAssertNil(request)
+            // when
+            let payload = [
+                "status": "disabled"
+            ]
+
+            let response = ZMTransportResponse(payload: payload as NSDictionary as ZMTransportData, httpStatus: 200, transportSessionError: nil)
+            self.sut.update(feature, with: response, downstreamSync: self.sut.fetchSingleConfigSync)
```
typfel, line 97: The test will be more resilient if you call `request.complete(with:)` instead of calling `self.sut.update(,with,downstreamSync:)`. That way it doesn't depend on how FeatureConfigRequestStrategy is implemented internally.

typfel (changes requested):

Sources/Request Strategies/Feature configurations/FeatureConfigRequestStrategyTests.swift
KaterinaWire: Done

PR #363, opened 2021-09-02
[Fix] Duplicate messages being sent

typfel:
### Issues

It been observed that messages were sometimes sent multiple times.

### Causes

In the case when a message has multiple dependencies which needs to be resolved before it can be sent it would be added to the list of "messages to send" once per dependency.

### Solutions

When a message has resolved its dependencies, clear all dependencies immediately.


johnxnguyen (approved):

KaterinaWire (approved):

typfel:
@zenkins release patch

PR #373, opened 2021-09-20
Fix: xcode 13 String Format Specifiers %u does not compile

billypchan:
## What's new in this PR?

Fix the compiler error on Xcode 13 that `%u` is not allowed for `unsigned` type. ( `imageOwnersThatNeedPreprocessing` is `NSMutableOrderedSet` and should not be casted to unsigned int)


KaterinaWire (approved):

David-Henner (approved):

billypchan:
@zenkins release patch

PR #374, opened 2021-09-20
[Feature] Integrate new conversation request strategy

typfel:
## What's new in this PR?

Adds `ConversationRequestStrategy` as a new strategy for synchronising conversations, which is federation aware. This replace the old `ZMConversationTranscoder`.

Improvements over `ZMConversationTranscoder`:
- Written Swift
- Decodes payloads using decodable structs.
- Payload processing logic is separated away from the strategy.
- Easy to remove legacy endpoint support when federation is deployed everywhere.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1236


billypchan (approved):

Sources/Request Strategies/Conversation/ConversationRequestStrategyTests.swift
```diff
@@ -0,0 +1,1111 @@
...
+            let payload: [String: Any] = [
+                "from": selfUser.remoteIdentifier!.transportString(),
+                "conversation": self.groupConversation!.remoteIdentifier!.transportString(),
+                "time": NSDate(timeIntervalSinceNow: 100).transportString(),
+                "data": [
+                    "target": userId.transportString(),
+                    "conversation_role": "new"
+                ],
+                "type": "conversation.member-update"
+            ]
```
billypchan, line 858: The repeated payload init code is hard to read the difference. How about make a factory method like `func payload(time: Date, .....)` ?

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -0,0 +1,790 @@
...
+    let conversationByQualifiedIDSync: IdentifierObjectSync<ConversationByQualifiedIDTranscoder>
+
+    let conversationByIDListTranscoder: ConversationByIDListTranscoder
+    let conversationByIDListSync: IdentifierObjectSync<ConversationByIDListTranscoder>
+
+    let conversationByQualifiedIDListTranscoder: ConversationByQualifiedIDListTranscoder
+    let conversationByQualifiedIDListSync: IdentifierObjectSync<ConversationByQualifiedIDListTranscoder>
+
+    var insertSync: ZMUpstreamInsertedObjectSync!
+    var modifiedSync: ZMUpstreamModifiedObjectSync!
```
billypchan, line 39: there two are assigned after self is created so have to be var optional. How about make it as `lazy var` to get rid of optional and make it get only?

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -0,0 +1,790 @@
...
+            conversationByQualifiedIDSync.sync(identifiers: identifiers)
+        } else {
+            conversationByIDSync.sync(identifiers: converations.compactMap(\.remoteIdentifier))
+        }
+    }
+
+    func fetchAllConversations() {
+        guard !isFetchingAllConversations else { return }
+
+        isFetchingAllConversations = true
```
billypchan, line 128: this bool looks like doing a job `dispatchSemaphone`.

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -0,0 +1,790 @@
...
+                    self?.conversationByIDListSync.sync(identifiers: conversationIDList.conversations)
+                case .failure:
+                    self?.syncProgress.failCurrentSyncPhase(phase: .fetchingConversations)
+                }
+            }
+        }
+    }
+
+    public var requestGenerators: [ZMRequestGenerator] {
+        if syncProgress.currentSyncPhase == .fetchingConversations {
```
billypchan, line 159: use switch?

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -0,0 +1,790 @@
...
+                .conversationMemberLeave,
+                .conversationMemberJoin,
+                .conversationRename,
+                .conversationMemberUpdate,
+                .conversationAccessModeUpdate,
+                .conversationMessageTimerUpdate,
+                .conversationReceiptModeUpdate,
+                .conversationConnectRequest
+        ]
+    }
```
billypchan, line 193:
```suggestion
    var eventsToProcess: [ZMUpdateEventType] = [.conversationCreate,
                .conversationDelete,
                .conversationMemberLeave,
                .conversationMemberJoin,
                .conversationRename,
                .conversationMemberUpdate,
                .conversationAccessModeUpdate,
                .conversationMessageTimerUpdate,
                .conversationReceiptModeUpdate,
                .conversationConnectRequest
        ]
```

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -0,0 +1,790 @@
...
+                let payloadAsDictionary = event.payload as? [String: Any],
+                let payloadData = try? JSONSerialization.data(withJSONObject: payloadAsDictionary, options: [])
+            else {
+                continue
+            }
+
+            switch event.type {
+            case .conversationCreate:
+                let conversationEvent = Payload.ConversationEvent<Payload.Conversation>(payloadData)
+                conversationEvent?.process(in: managedObjectContext, originalEvent: event)
```
billypchan, line 210: we can save lines by moving `conversationEvent?.process(in: managedObjectContext, originalEvent: event)` outside the switch. And put `let conversationEvent: SomeType` before the switch

billypchan (commented):
clean up suggestions

KaterinaWire (approved):

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
typfel: I don't need any semaphores since it's guaranteed to happen on the same thread. I don't be wait on `fetchAllConversations` if it's already in progress since that would create deadlock in the sync context.

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
typfel: I only care about one case so I think a `switch` would be more verbose.

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
typfel: I can't move it out of the switch since they all have different types.

billypchan (approved):

typfel:
@zenkins release major

PR #379, opened 2021-10-01
[Fix] 1-1 Calls sometimes not connecting 

typfel:
### Issues

It's been observed that 1-1 Calls are sometimes not connecting.

### Causes

If the call message didn't need any further requests for establishing sessions, the SETUP message wouldn't be sent until another event occurred in the application. This behaviour comes from the fact that calling messages are not persisted in Core Data which is what normally triggers the operation loop to send new requests.

### Solutions

Always explicitly tell the operation loop that there's new requests available after scheduling a message to be sent.

## Notes

This bug was introduced by https://github.com/wireapp/wire-ios-request-strategy/pull/361


billypchan (approved):

David-Henner (approved):

typfel:
@zenkins release patch

PR #381, opened 2021-10-08
[Fix] Don't crash when to field is nil due to connection being cancelled

typfel:
### Issues

The app would crash when syncing the user profiles of your connections.

### Causes

If a connection is cancelled the `to` field is nulled out and `connections.map(\.to)` would crash due to force unwrapping.

### Solutions

Use `compactMap` instead.


David-Henner (approved):

typfel:
@zenkins release patch

PR #382, opened 2021-10-11
[Feature] Federation aware connections

typfel:
## What's new in this PR?

This is part of a series of PRs which refactor connections and make them federation aware.

- `ConnectionRequestStrategy` replaces `ZMConnectionTranscoder`
- ZMConnection are no longer inserted locally in UI context to be synchronised by SE context at a later point. This behaviour is now implemented  by the `ConnectToUserAction` which means that `ZMConnection` objects are only inserted in SE context which avoid many possible race conditions.
- ZMConnection no longer synchronises any state by observing the `status` field. This behaviour is now implemented by the `UpdateConnectionAction`.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1241


Sources/Request Strategies/Connection/ConnectionRequestStrategy.swift
```diff
@@ -0,0 +1,290 @@
...
+
+                    if !connectionList.hasMore {
+                        self?.syncProgress.finishCurrentSyncPhase(phase: .fetchingConnections)
+                        self?.isFetchingAllConnections = false
+                    }
+                case .failure:
+                    self?.syncProgress.failCurrentSyncPhase(phase: .fetchingConnections)
+                }
+            }
+        }
```
David-Henner, line 129: the two fetch completion blocks look identical, is there a way to reduce the duplication here ?

Sources/Request Strategies/Connection/ConnectionRequestStrategy.swift
```diff
@@ -0,0 +1,290 @@
...
+        } else {
+            if let userID = object.to.remoteIdentifier {
+                connectionByIDSync.sync(identifiers: Set(arrayLiteral: userID))
+            }
+        }
+    }
+
+    func cancel(_ object: ZMConnection) {
+        
+    }
```
David-Henner, line 167: is implementation missing?

Sources/Request Strategies/Connection/ConnectionRequestStrategy.swift
```diff
@@ -0,0 +1,290 @@
...
+                conversationEvent?.process(in: managedObjectContext)
+            default:
+                break
+            }
+        }
+    }
+
+}
+
+class ConnectionByIDTranscoder: IdentifierObjectSyncTranscoder {
```
David-Henner, line 195: will this `ConnectionByIDTranscoder` be completely replaced by `ConnectionByQualifiedIDTranscoder` at some point?

Sources/Payloads/Processing/Payload+Processing.swift
```diff
@@ -0,0 +1,231 @@
...
+                                                                          selfClient: selfClient)
+                } else {
+                    missingClient.markClientAsInvalidAfterFailingToRetrievePrekey(selfClient: selfClient)
+                }
+
+
+            }
+        }
+
+        let hasMoreMissingClients = (selfClient.missingClients?.count ?? 0) > 0
```
billypchan, line 100:
```suggestion
        let hasMoreMissingClients = selfClient.missingClients?.isEmpty == false
```

PR #401, opened 2021-11-03
Fix: Handle conflicting domains FS-119

typfel:
### Issues

User and conversations are created with conflicting domains. 

### Causes

See https://github.com/wireapp/wire-ios-data-model/pull/1254 for context

### Solutions

Don't assign the domain if federation is not enabled.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/1254


Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
```diff
@@ -166,7 +166,7 @@ extension Payload.Conversation {
 
         conversation.conversationType = .group
         conversation.remoteIdentifier = conversationID
-        conversation.domain = qualifiedID?.domain
+        conversation.domain = context.zm_isFederationEnabled ? qualifiedID?.domain : nil
```
billypchan, line 169: repeated code can be cleaned up by write a private method with context and qualifiedID as parameter.

billypchan (commented):
code clean up suggestion

Sources/Payloads/Processing/PayloadProcessing+ConversationTests.swift
```diff
@@ -37,6 +38,22 @@ class PayloadProcessing_ConversationTests: MessagingTestBase {
         }
     }
 
+    func testUpdateOrCreateConversation_Group_DoesntUpdatesQualifiedID_WhenFederationIsDisabled() throws {
+        syncMOC.performGroupedBlockAndWait {
+            // given
+            self.syncMOC.zm_isFederationEnabled = false
+            let qualifiedID = self.groupConversation.qualifiedID!
+            let conversation = Payload.Conversation(qualifiedID: qualifiedID, id: qualifiedID.uuid)
```
David-Henner, line 46: wouldn't this update the qualified ID to the same values as it already has? Or am I missing something?

typfel: That's true. But consider:
- Same code path is used to insert a new conversation if doesn't exist
- We might already have the conversation locally but might be missing the domain field.

Sources/Payloads/Processing/PayloadProcessing+ConversationTests.swift
```diff
@@ -37,6 +38,22 @@ class PayloadProcessing_ConversationTests: MessagingTestBase {
...
+
+            // when
+            conversation.updateOrCreate(in: self.syncMOC)
+
+            // then
+            XCTAssertEqual(self.groupConversation.remoteIdentifier, qualifiedID.uuid)
+            XCTAssertEqual(self.groupConversation.domain, qualifiedID.domain)
```
typfel, line 53: This should be: `XCTAssertNil(self.groupConversation.domain)`

I'll fix this.

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
typfel: Yes, we could but I think it would actually make the code a harder to read.

Sources/Payloads/Processing/PayloadProcessing+ConversationTests.swift
typfel: Done

billypchan (approved):

typfel:
release major

PR #414, opened 2021-11-11
Fix: Sending via share extension doesn't work

typfel:
### Issues

 Sending via share extension doesn't work

### Causes

Share extension attempts to use federation endpoint which require domains but they aren't available.

### Solutions

Disable federation endpoints by default.

## Notes

Future-work: respect the federation feature flag in the share extension.


KaterinaWire (approved):

billypchan (approved):

typfel:
release patch

PR #442, opened 2021-12-02
chore: Fix SwiftLint warnings in Source Part 3

agisilaos:
----

- The **PR Title**
  - [ ] contains a reference JIRA issue number like `SQPIT-764`

- The **PR Description**
----

# What's new in this PR?

In this PR I fix warnings related to the SwiftLint rules below and a warning from the Swift Compiler:

- **Vertical Parameter Alignment Violation**: Function parameters should be aligned vertically if they're in multiple lines in a declaration. `(vertical_parameter_alignment)`
- **Computed Accessors Order Violation**: Computed properties should declare first the getter and then the setter. `(computed_accessors_order)`
- **Operator Function Whitespace Violation**: Operators should be surrounded by a single whitespace when defining them. `(operator_whitespace)`
- **Duplicate Imports Violation**: Imports should be unique. `(duplicate_imports)`
- **Colon Violation**: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. `(colon)`
- **Implicit Getter Violation**: Computed read-only properties should avoid using the get keyword. `(implicit_getter)`
- **Is Disjoint Violation**: Prefer using `Set.isDisjoint(with:)` over `Set.intersection(_:).isEmpty`.`(is_disjoint)`
- Using 'class' keyword to define a class-constrained protocol is deprecated; use 'AnyObject' instead
- **Empty Enum Arguments Violation**: Arguments can be omitted when matching enums with associated values if they are not used. `(empty_enum_arguments)`

----
##### References


Sources/Protocols/OTREntity.swift
```diff
@@ -106,6 +106,7 @@ extension OTREntity {
         if let selfClient = ZMUser.selfUser(in: context).selfClient(),
            let missingClients = selfClient.missingClients, missingClients.count > 0 {
             // Don't block sending of messages if they are not affected by the missing clients
+            // Is Disjoint Violation: Prefer using `Set.isDisjoint(with:)` over `Set.intersection(_:).isEmpty`. (is_disjoint)
             if !missingClients.intersection(recipientClients).isEmpty {
```
agisilaos, line 110: Not sure how to fix this warning so any help is much appreciated.

Sources/Protocols/OTREntity.swift
```diff
@@ -106,7 +106,8 @@ extension OTREntity {
         if let selfClient = ZMUser.selfUser(in: context).selfClient(),
            let missingClients = selfClient.missingClients, missingClients.count > 0 {
             // Don't block sending of messages if they are not affected by the missing clients
-            if !missingClients.intersection(recipientClients).isEmpty {
+            // Is Disjoint Violation: Prefer using `Set.isDisjoint(with:)` over `Set.intersection(_:).isEmpty`. (is_disjoint)
+            if !missingClients.isDisjoint(with: recipientClients) {
```
agisilaos, line 110: Not sure if this is the correct fix for the warning. Let me know what you think.

johnxnguyen: It's correct, but no need for the additional comment.

johnxnguyen (changes requested):

johnxnguyen (approved):

David-Henner (approved):

PR #481, opened 2022-01-27
feat: Separate access controls for guests ands services SQSERVICES-1052

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1052" title="SQSERVICES-1052" target="_blank">StorySQSERVICES-1052</a>  [iOS] Separate access controls for guests and services
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

With this PR I continue the support for `access_role_v2` in Request Strategy as well. We may need to add more tests but that's up for discussion. cc: @johnxnguyen 

### Dependencies (Optional)

- https://github.com/wireapp/wire-ios-data-model/pull/1322
- https://github.com/wireapp/wire-ios/pull/5516
- https://github.com/wireapp/wire-ios-sync-engine/pull/1644

----

##### References


Sources/Request Strategies/Conversation/ConversationRequestStrategyTests.swift
```diff
@@ -483,10 +483,10 @@ class ConversationRequestStrategyTests: MessagingTestBase {
         self.syncMOC.performAndWait {
 
             let newAccessMode = ConversationAccessMode(values: ["code", "invite"])
-            let newAccessRole = ConversationAccessRole.team
+            let newAccessRole: Set<ConversationAccessRoleV2> = [.teamMember, .guest]
```
KaterinaWire, line 486:
```suggestion
            let newAccessRoles: Set<ConversationAccessRoleV2> = [.teamMember, .guest]
```

Sources/Payloads/Processing/PayloadProcessing+ConversationTests.swift
```diff
@@ -203,19 +203,19 @@ class PayloadProcessing_ConversationTests: MessagingTestBase {
...
             let accessMode: ConversationAccessMode = .allowGuests
-            let accessRole: ConversationAccessRole = .team
+            let accessRole: Set<ConversationAccessRoleV2> = [.teamMember]
             let qualifiedID = self.groupConversation.qualifiedID!
 
             let conversationPayload = Payload.Conversation(qualifiedID: qualifiedID,
                                                            type: BackendConversationType.group.rawValue,
                                                            access: accessMode.stringValue,
-                                                           accessRole: accessRole.rawValue)
+                                                           accessRoleV2: accessRole.map {$0.rawValue})
```
johnxnguyen, line 212:
```suggestion
                                                           accessRoleV2: accessRole.map(\.rawValue))
```

johnxnguyen: I would also expect to see some tests that handle the fallback code. For instance:
- given access_role_v2 is present in the payload, then we use it.
- given access_role_v2 is not present, then we use access_role

Sources/Request Strategies/Conversation/ConversationRequestStrategyTests.swift
```diff
@@ -495,15 +495,15 @@ class ConversationRequestStrategyTests: MessagingTestBase {
                                          timestamp: Date(),
                                          dataPayload: [
                                             "access": newAccessMode.stringValue,
-                                            "access_role": newAccessRole.rawValue
+                                            "access_role_v2": newAccessRole.map { $0.rawValue }
```
johnxnguyen, line 498:
```suggestion
                                            "access_role_v2": newAccessRole.map(\.rawValue)
```

johnxnguyen (approved):

Sources/Request Strategies/Conversation/ConversationRequestStrategyTests.swift
KaterinaWire, line 486: Why did you add `.guest` option if the original test was only for team members?

agisilaos: So they're not equal, in line 486. If I leave just the `.teamMember` test fails in line 486 with an error:

`XCTAssertNotEqual failed: ("[WireDataModel.ConversationAccessRoleV2.teamMember]") is equal to ("[WireDataModel.ConversationAccessRoleV2.teamMember]")

johnxnguyen: I guess it's irrelevant what the new value is, only that it's different from the current value before the update event is processed, and then equal after the update event is processed.

agisilaos: That's what I thought, after running the test with breakpoints and checking again the values to make sure I didn't miss something.

KaterinaWire: I agree that `updateEvent(...)` method is working fine and we check it in this method. But if we need to change the initial value, it looks like we've broken something and it might affect something else. That's why I was wondering why we need to change this part. Currently the `self.groupConversation.accessRoles` is ConversationAccessRoleV2**.teamMember**, but it looks like the old `self.groupConversation.accessRole` had a different value and we don't know why. 
Anyway, I don't mind merging these changes and maybe we'll find out about it later or it won't affect anything.

KaterinaWire (approved):

Sources/Request Strategies/Conversation/ConversationRequestStrategyTests.swift
johnxnguyen: Good point. I suppose then that the default value in `groupConversation` has changed. @agisilaos do you know why that may be?

agisilaos: Hmmm. Not sure. 🤔

johnxnguyen: ah I know why. It's because previously `accessRole` was initially `nil`, but `accessRoles` has a default value of `[.teamMember]`. I think it's fine, it's a sensible default value.

agisilaos: It was nil indeed, right! The default value for accessRoles as per the spec and what we already discussed is `[.teamMember]`

agisilaos: Should I go ahead and merge it then? @KaterinaWire Any objections or questions?

KaterinaWire: I don't have objections

PR #500, opened 2022-03-18
feat: consider api version FS-451

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR we replace the use of existing federation flags with the api version. The federation flags were used to determine whether to make requests for the federated endpoints. This flag is replaced as follows: an api version of `v0` corresponds to the federation flag being `false`, and a version of `v1` corresponds to the flag being `true`.

Most changes are fairly straightforward, but those involving the object identifier syncs are more involved. Here we could not always rely on the api version being passed into the current scope (such as when reacting to changed object notifications). In such cases, we obtain the api version from the `APIVersion.current` helper.

In some cases we were attempting to speak to federated endpoints and would fallback to legacy endpoints if it failed. This logic has been removed since we now rely on the api version. If we're on version 1, then we should assume the endpoint exists.

There are significant changes to update the tests. I feel that there needs to be clearer separation in tests when testing different api versions. This may come in a separate PR.

### Dependencies 

Needs releases with:

- https://github.com/wireapp/wire-ios-transport/pull/247

### Testing

Updated tests, but didn't need to add any new ones.

----

----

----
##### References


Sources/Object Syncs/Helpers/ProteusMessageSyncTests.swift
```diff
@@ -126,39 +125,20 @@ class ProteusMessageSyncTests: MessagingTestBase {
...
-            sut.nextRequest(for: .v0)?.complete(with: ZMTransportResponse(payload: payloadAsString as ZMTransportData, httpStatus: payload.code, transportSessionError: nil, apiVersion: APIVersion.v0.rawValue))
-        }
-        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
-
-        syncMOC.performGroupedBlockAndWait { [self] in
-            // then
-            XCTAssertEqual(sut.nextRequest(for: .v0)?.path, legacyEndpoint)
-        }
-    }
-
```
johnxnguyen, line 147: This tests the fallback mechanism, which isn't needed anymore.

Sources/Request Strategies/Assets/Helpers/AssetDownloadRequestFactory.swift
```diff
@@ -19,25 +19,22 @@
...
-        if useFederationEndpoint {
-            guard let domain = domain else { return nil }
 
-            path = "/assets/v4/\(domain)/\(key)"
-        } else {
+        switch apiVersion {
+        case .v0:
             path = "/assets/v3/\(key)"
+        case .v1:
+            guard let domain = domain else { return nil }
```
David-Henner, line 31: Where does the domain come from? Should we fall back on the one from `APIVersion` if it's nil?

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -141,21 +133,28 @@ public class ConversationRequestStrategy: AbstractRequestStrategy, ZMRequestGene
...
+        switch apiVersion {
+        case .v0:
             conversationByIDSync.sync(identifiers: converations.compactMap(\.remoteIdentifier))
+
+        case .v1:
+            if let qualifiedIDs = converations.qualifiedIDs {
+                conversationByQualifiedIDSync.sync(identifiers: qualifiedIDs)
+            } else {
+                // Fallback to unqualified ids.
+                conversationByIDSync.sync(identifiers: converations.compactMap(\.remoteIdentifier))
```
johnxnguyen, line 152: Here I persevered the existing logic, but not sure if this is what we need to do.

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -276,15 +277,33 @@ extension ConversationRequestStrategy: KeyPathObjectSyncTranscoder {
...
+        switch apiVersion {
+        case .v0:
+            guard let identifier = object.remoteIdentifier else { return }
+            synchronize(unqualifiedID: identifier)
+
+        case .v1:
+            if let qualifiedID = object.qualifiedID {
+                synchronize(qualifiedID: qualifiedID)
+            } else if let identifier = object.remoteIdentifier {
+                synchronize(unqualifiedID: identifier)
```
johnxnguyen, line 292: Same here, we have a fallback mechanism but not sure if we need to keep this.

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
David-Henner: If federation is disabled, we won't have qualified ids for users and conversation. So it makes sense to have a fallback in that case. We may also want to explicitly check the federation flag to make it clear that when disabled we shouldn't use qualified ids. 
If federation is enabled, I assume that the backend will expect qualified ids, so I don't know if it makes a lot of sense to fallback

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
David-Henner: Same comment here

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -444,16 +463,17 @@ extension ConversationRequestStrategy: ZMUpstreamTranscoder {
...
+            case .v0:
+                request = ZMTransportRequest(path: "/conversations/\(conversationID)",
                                              method: .methodPUT,
                                              payload: payloadAsString as ZMTransportData?,
                                              apiVersion: apiVersion.rawValue)
-            } else {
-                request = ZMTransportRequest(path: "/conversations/\(conversationID)",
+
+            case .v1:
+                guard let domain = conversation.domain else { return nil }
```
David-Henner, line 475: if federation is not enabled we need to fallback to the domain from `APIVersion` if we want to use the federation endpoints

Sources/Request Strategies/User Clients/MissingClientsRequestStrategy.swift
```diff
@@ -131,19 +118,25 @@ public final class MissingClientsRequestStrategy: AbstractRequestStrategy, ZMUps
...
                     return false
                 }
 
                 return prekeys.establishSessions(with: selfClient, context: managedObjectContext)
-            } else {
+
+            case .v1:
                 guard let rawData = response.rawData,
-                      let prekeys = Payload.PrekeyByUserID(rawData),
+                      let prekeys = Payload.PrekeyByQualifiedUserID(rawData),
```
David-Henner, line 139: also here, not sure if we can use qualified ids when federation is disabled

Sources/Request Strategies/User/UserProfileRequestStrategy.swift
```diff
@@ -104,10 +83,24 @@ public class UserProfileRequestStrategy: AbstractRequestStrategy, IdentifierObje
...
+        guard !users.isEmpty else { return }
+
+        switch apiVersion {
+        case .v0:
+            userProfileByID.sync(identifiers: users.compactMap(\.remoteIdentifier))
+
+        case .v1:
+            if let qualifiedUserIDs = users.qualifiedUserIDs {
+                userProfileByQualifiedID.sync(identifiers: qualifiedUserIDs)
+            }
```
David-Henner, line 100: also here we would need to fallback

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
johnxnguyen: I lost the other comment but if I remember correctly.... here it's possible there is no qualified id (because federation is disabled), in which case we should use the local domain. This means we would wouldn't ever need to call `synchronise(unqualifiedID:)` right?

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -141,21 +133,28 @@ public class ConversationRequestStrategy: AbstractRequestStrategy, ZMRequestGene
...
+        case .v0:
+            conversationByIDSync.sync(identifiers: conversations.compactMap(\.remoteIdentifier))
+
+        case .v1:
+            if let qualifiedIDs = conversations.qualifiedIDs {
+                conversationByQualifiedIDSync.sync(identifiers: qualifiedIDs)
+            } else {
+                // Fallback to unqualified ids.
+                conversationByIDSync.sync(identifiers: conversations.compactMap(\.remoteIdentifier))
+            }
```
johnxnguyen, line 153: Here we shouldn't fallback to unqualified ids, but instead always have a qualified id (using the local domain if needed)... right?

Sources/Request Strategies/Assets/Helpers/AssetDownloadRequestFactory.swift
johnxnguyen: It comes from the protobuf object, I guess it won't hurt to fallback to the local domain.

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
johnxnguyen: Fixed

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
johnxnguyen: Fixed

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -476,16 +496,17 @@ extension ConversationRequestStrategy: ZMUpstreamTranscoder {
...
+            case .v0:
+                request = ZMTransportRequest(path: "/conversations/\(conversationID)/self",
                                              method: .methodPUT,
                                              payload: payloadAsString as ZMTransportData?,
                                              apiVersion: apiVersion.rawValue)
-            } else {
-                request = ZMTransportRequest(path: "/conversations/\(conversationID)/self",
+
+            case .v1:
+                guard let domain = conversation.domain else { return nil }
```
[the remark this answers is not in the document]
johnxnguyen: Fixed

Sources/Request Strategies/User Clients/MissingClientsRequestStrategy.swift
```diff
@@ -115,10 +100,12 @@ public final class MissingClientsRequestStrategy: AbstractRequestStrategy, ZMUps
...
         let request: ZMUpstreamRequest?
-        if isFederationEndpointAvailable {
-            request = requestsFactory.fetchPrekeysFederated(for: missing, apiVersion: apiVersion)
-        } else {
+
+        switch apiVersion {
+        case .v0:
             request = requestsFactory.fetchPrekeys(for: missing, apiVersion: apiVersion)
+        case .v1:
+            request = requestsFactory.fetchPrekeysFederated(for: missing, apiVersion: apiVersion)
```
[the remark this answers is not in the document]
johnxnguyen: Fixed in 5afb9d83c0aa250aee7ca314d737d8b82365e804

Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -823,3 +841,19 @@ class ConversationByQualifiedIDListTranscoder: IdentifierObjectSyncTranscoder {
...
+
+    func fallbackQualifiedIDs(localDomain: String) -> [QualifiedID] {
+        return compactMap { converation in
+            if let qualifiedID = converation.qualifiedID {
+                return qualifiedID
+            } else if let identifier = converation.remoteIdentifier {
+                return QualifiedID(uuid: identifier, domain: localDomain)
+            } else {
+                return nil
+            }
```
David-Henner, line 855:
```suggestion
        return compactMap { conversation in
            if let qualifiedID = conversation.qualifiedID {
                return qualifiedID
            } else if let identifier = conversation.remoteIdentifier {
                return QualifiedID(uuid: identifier, domain: localDomain)
            } else {
                return nil
            }
```

agisilaos (approved):

PR #513, opened 2022-05-02
feat: process v1 payload of missing clients response FS-635

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-635" title="FS-635" target="_blank">TaskFS-635</a>  [iOS] Support parsing of federation enabled v1 client discovery payload for legal hold
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

The `VerifyLegalHoldRequestStrategy` was missing support to process the payload from the federation-enabled missing clients endpoint. This endpoint should be requested when the api version is `v1`. In the PR https://github.com/wireapp/wire-ios-request-strategy/pull/505, we forced using the legacy endpoint as a temporary solution. However, we need to use the new endpoint, since we will need to support legal hold for federation at some point.

**legacy endpoint**
`/conversations/{cnv}/otr/messages`

**new endpoint (federation-enabled)**
`/conversations/{cnv_domain}/{cnv}/proteus/messages`

The new endpoint comes with a different payload. The `Codable` struct `Payload.MessageSendingStatus` already exists and allows us to decode it.

#### Changes

- Added a method on `Payload.MessageSendingStatus` to map the payload to a dictionary of users by client list.
- Updated the response processing to account for the different api versions and extract a generic client list from each payload, which then can be processed by the existing code, thus reducing duplication of the logic.

#### Test Coverage (Optional)

Added tests to: 
- `VerifyLegalHoldRequestStrategyTests` for `v1`
- `Payload+MessageSendingStatusTests`. Testing that the new payload mapping method works as intended.

#### How to Test

Once merged with the UI, it can be tested by verifying the list of users under legal hold in a conversation is correct

----

----

----
##### References


Sources/Protocols/OTREntity.swift
```diff
@@ -121,29 +121,48 @@ extension OTREntity {
         return nil
     }
 
-    public func parseEmptyUploadResponse(_ response: ZMTransportResponse, in conversation: ZMConversation, clientRegistrationDelegate: ClientRegistrationDelegate) -> ZMConversationRemoteClientChangeSet {
```
David-Henner, line 124: removed `public` because it's only used in request strategy

Sources/Payloads/Processing/PayloadProcessing+MessageSendingStatus.swift
```diff
@@ -60,4 +64,26 @@ extension Payload.MessageSendingStatus {
         return !missingClients.isEmpty
     }
 
+    func missingClientListByUser(context: NSManagedObjectContext) -> Payload.ClientListByUser {
+
+        let clientListByUserTuples = missing.flatMap { (domain, userClientsByUserID) in
+            userClientsByUserID.mappingUserIDsToUsers(in: context, domain: domain)
+        }
+
+        return Payload.ClientListByUser(clientListByUserTuples, uniquingKeysWith: +)
```
johnxnguyen, line 73: It's a little hard to understand the transformation that is happening here. This is how I understand it:

- `missing` is a dictionary `String[: [UUID: [String]]]` that maps the domain to another dictionary that maps user ids to list of client ids.
- We want to obtain another dictionary `[ZMUser: [String]]` that maps a user to a list of client ids. 
- To achieve this, we need to create the `ZMUser` objects using the `domain` and the user id and then flatten all the results.

If I understand that correctly, then I'd suggest to rename some things to make it a bit clearer. For instance:

```swift
        let clientIDsByUser = missing.flatMap { domain, clientIDsByUserID in
            clientIDsByUserID.materializingUsers(withDomain: domain, in: context)
        }

        return Payload.ClientListByUser(clientIDsByUser, uniquingKeysWith: +)
```

Sources/Payloads/Processing/PayloadProcessing+MessageSendingStatus.swift
```diff
@@ -60,4 +64,26 @@ extension Payload.MessageSendingStatus {
...
+extension Payload.ClientListByUserID {
+
+    func mappingUserIDsToUsers(in context: NSManagedObjectContext, domain: String?) -> [ZMUser: [String]] {
+
+        return reduce(into: Payload.ClientListByUser()) { (result, tuple: (userID: String, clientIDs: [String])) in
+            guard let userID = UUID(uuidString: tuple.userID) else { return }
+            let user = ZMUser.fetchOrCreate(with: userID, domain: domain, in: context)
+            result[user] = tuple.clientIDs
+        }
+    }
```
johnxnguyen, line 87: The name is a bit confusing because the word `mapping` can mean both "converting" (I think the intended meaning here) and "paring" (as in.. creating a dictionary). I suggest something like `materializingUsers` to indicate we want to convert user ids to users.

Also, if I understand correctly the transformation of the keys from `UUID` to `ZMUser`, another way you could express this method is:

```swift
func materializingUsers(withDomain domain: String?, in context: NSManagedObjectContext) -> [ZMUser: [String]] {
    let tuples = compactMap { userID, clientIDs in
        guard let userID = UUID(uuidString: userID) else { return nil }
        let user = ZMUser.fetchOrCreate(with: userID, domain: domain, in: context)
        return (user, clientIDs)
    }

    return Dictionary(uniqueKeysWithValues: tuples)
}
```

Not sure if that is more readable... but thought I'd share it anyway :)

johnxnguyen (approved):

agisilaos (approved):

David-Henner:
release major

zenkins:
Hello, @David-Henner! Merging PR for command **release major** for PR #**513**. Will create a **major** release after merged.
Next version is 266.0.0.

PR #528, opened 2022-06-16
feat(mls): key package count action for self user FS-695

ghuMustafa:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-695" title="FS-695" target="_blank">Sub-taskFS-695</a>  [iOS] Create `CountSelfMLSKeyPackagesAction`
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- Created an action `CountSelfMLSKeyPackagesAction` for counting the number or remaining key packages for the self user.
- Created an action handler `CountSelfMLSKeyPackagesActionHandler` for `CountSelfMLSKeyPackagesAction`.
- Added the action handler to the `MLSRequestStrategy`.

### Testing

- [x] Created tests for `CountSelfMLSKeyPackagesActionHandler`

----

----

----
##### References

### DIFF
--- a/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
+++ b/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
@@ -1,0 +1,61 @@
+//
+// Wire
+// Copyright (C) 2022 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public final class CountSelfMLSKeyPackagesAction: EntityAction {
+
+    // MARK: - Types
+
+    public typealias Result = Int
+
+    public enum Failure: Error {
+
+        case clientNotFound
+        case malformedResponse
+        case unknown(status: Int)
+
+        public var errorDescription: String? {
+            switch self {
+            case .clientNotFound:
+                return "Client not found"
+            case .malformedResponse:
+                return "Malformed response"
+            case .unknown(let status):
+                return "Unknown error (response status: \(status))"
+            }
+        }
+
+        public var safeForLoggingDescription: String {
+            return errorDescription ?? ""
+        }
+    }
+
+    // MARK: - Properties
+
+    public var resultHandler: ResultHandler?
+    public var clientID: String
+
+    // MARK: - Life cycle
+
+    init(clientID: String,
+         resultHandler: ResultHandler? = nil) {
+        self.clientID = clientID
+        self.resultHandler = resultHandler
+    }
+}
--- a/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
+++ b/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
@@ -29,6 +29,7 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
         case clientNotFound
         case malformedResponse
         case unknown(status: Int)
+        case invalidClientID
 
         public var errorDescription: String? {
             switch self {
@@ -38,6 +39,8 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
                 return "Malformed response"
             case .unknown(let status):
                 return "Unknown error (response status: \(status))"
+            case .invalidClientID:
+                return "Invalid clientID"
             }
         }
     }
--- a/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
+++ b/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
@@ -40,10 +40,6 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
                 return "Unknown error (response status: \(status))"
             }
         }
-
-        public var safeForLoggingDescription: String {
-            return errorDescription ?? ""
-        }
     }
 
     // MARK: - Properties
--- a/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
+++ b/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
@@ -28,8 +28,9 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
 
         case clientNotFound
         case malformedResponse
-        case unknown(status: Int)
+        case unknown(status: Int?)
         case invalidClientID
+        case apiVersionNotSupported
 
         public var errorDescription: String? {
             switch self {
--- a/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
+++ b/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
@@ -28,9 +28,9 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
 
         case clientNotFound
         case malformedResponse
-        case unknown(status: Int)
         case invalidClientID
-        case apiVersionNotSupported
+        case endpointNotAvailable
+        case unknown(status: Int)
 
         public var errorDescription: String? {
             switch self {
@@ -38,12 +38,12 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
                 return "Client not found"
             case .malformedResponse:
                 return "Malformed response"
-            case .unknown(let status):
-                return "Unknown error (response status: \(status))"
+            case .endpointNotAvailable:
+                return "End point not available"
             case .invalidClientID:
                 return "Invalid clientID"
-            case .apiVersionNotSupported:
-                return "Api version not supported"
+            case .unknown(let status):
+                return "Unknown error (response status: \(status))"
             }
         }
     }
--- a/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
+++ b/Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
@@ -28,7 +28,7 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
 
         case clientNotFound
         case malformedResponse
-        case unknown(status: Int?)
+        case unknown(status: Int)
         case invalidClientID
         case apiVersionNotSupported
 
@@ -42,6 +42,8 @@ public final class CountSelfMLSKeyPackagesAction: EntityAction {
                 return "Unknown error (response status: \(status))"
             case .invalidClientID:
                 return "Invalid clientID"
+            case .apiVersionNotSupported:
+                return "Api version not supported"
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
```diff
@@ -0,0 +1,61 @@
...
+            case .clientNotFound:
+                return "Client not found"
+            case .malformedResponse:
+                return "Malformed response"
+            case .unknown(let status):
+                return "Unknown error (response status: \(status))"
+            }
+        }
+
+        public var safeForLoggingDescription: String {
```
johnxnguyen, line 44: This is only used if you want to be able to publicly log the value of this type. In this case, I don't think we want to (we limit public logs to only essential things). So I would remove this one.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
```diff
@@ -0,0 +1,68 @@
...
+
+import Foundation
+
+class CountSelfMLSKeyPackagesActionHandler: ActionHandler<CountSelfMLSKeyPackagesAction> {
+
+    // MARK: - Methods
+
+    override func request(
+        for action: ActionHandler<CountSelfMLSKeyPackagesAction>.Action,
+        apiVersion: APIVersion) -> ZMTransportRequest? {
```
johnxnguyen, line 27: In the method below you are formatting it like follows:

```suggestion
    override func request(
        for action: ActionHandler<CountSelfMLSKeyPackagesAction>.Action,
        apiVersion: APIVersion
    ) -> ZMTransportRequest? {
```

I think we should be consistent and change it here.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
```diff
@@ -0,0 +1,68 @@
...
+
+            return ZMTransportRequest(
+                path: "/mls/key-package/self/\(action.clientID)/count",
```
johnxnguyen, line 30: We should guard that the client ID is not empty, because it if is, then it would create an invalid path. In which case we'd need to notify the action failure and return nil. Or do you think it's better to do the validation when creating the action?

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
```diff
@@ -0,0 +1,68 @@
...
+                method: .methodGET,
+                payload: nil,
+                apiVersion: apiVersion.rawValue
```
johnxnguyen, line 33: We may also need to check the api version to see if this endpoint is only available from a certain version or not.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandlerTests.swift
```diff
@@ -0,0 +1,125 @@
...
+        // Expectation
+        let didFail = expectation(description: "didFail")
+
+        action.onResult { result in
+            guard case .failure(.malformedResponse) = result else { return }
+            didFail.fulfill()
+        }
+
+        // When
+        sut.handleResponse(response(payload: nil, status: 200), action: action)
```
johnxnguyen, line 81: To help indicate what is causing the malformed response, you could store pass in a variable like this:

```swift
let invalidPayload: ZMTransportData? = nil
sut.handleResponse(response(payload: invalidPayload, status: 200), action: action)
```

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
David-Henner: Do we already know if the MLS endpoints will only be available from `v1`?

ghuMustafa: Form where we can check Api version support? Is it available on swagger?

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: Definitely, its more clean and easy to read.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesAction.swift
ghuMustafa: Yes then it should't be needed here. Removed

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
David-Henner: I'm not seeing this on swagger, not sure where we can check.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandlerTests.swift
ghuMustafa: Yes I have specify type instead of just passing nil.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: Yes! we should guard `client ID` here and also while creating the action for double check.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
johnxnguyen: We'll have to check with the backend team directly.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
johnxnguyen: In addition to returning `nil`, I think it'd be good to also notify a failure on the action's result handler, so that we can surface the error.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
```diff
@@ -24,7 +24,12 @@ class CountSelfMLSKeyPackagesActionHandler: ActionHandler<CountSelfMLSKeyPackage
...
         for action: ActionHandler<CountSelfMLSKeyPackagesAction>.Action,
-        apiVersion: APIVersion) -> ZMTransportRequest? {
+        apiVersion: APIVersion
+    ) -> ZMTransportRequest? {
+
+        guard !action.clientID.isEmpty else {
+            return nil
+        }
 
             return ZMTransportRequest(
```
johnxnguyen, line 34: This method needs re-indentation.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: Yes I have also added failure notification.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: Yes! MLS endpoints will only be available from v1. Added.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: fixed.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    override func request(
+        for action: ActionHandler<CountSelfMLSKeyPackagesAction>.Action,
+        apiVersion: APIVersion
+    ) -> ZMTransportRequest? {
+
+        var action = action
+
+        guard apiVersion > .v0 else {
+            action.notifyResult(.failure(.apiVersionNotSupported))
```
johnxnguyen, line 33: Perhaps the error should be called `endpointNotAvailable`, since it's only available for v1 and above.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
```diff
@@ -0,0 +1,83 @@
...
+            return nil
+        }
+
+        guard !action.clientID.isEmpty else {
+            action.notifyResult(.failure(.clientNotFound))
```
johnxnguyen, line 38: I suggest to create a new error  called `invalidClientID`.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandlerTests.swift
```diff
@@ -0,0 +1,127 @@
...
+@testable import WireRequestStrategy
+
+class CountSelfMLSKeyPackagesActionHandlerTests: MessagingTestBase {
+
+    let clientID = "clientID"
+    let requestPath = "/v1/mls/key-package/self/clientID/count"
+
+    typealias Payload = CountSelfMLSKeyPackagesActionHandler.ResponsePayload
+
+    func test_sut_GeneratesValidRequest() throws {
```
johnxnguyen, line 29: We should also test the sad paths, such as no request (and action failure) if:
1. the api version is v0
2. the clientID is empty

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandlerTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        // When
+        let invalidPayload: ZMTransportData? = nil
+        sut.handleResponse(response(payload: invalidPayload, status: 200), action: action)
+
+        // Then
+        XCTAssert(waitForCustomExpectations(withTimeout: 0.5))
+    }
+
+    func test_sut_HandlesResponse_UnknownError() throws {
```
johnxnguyen, line 88: I think we're also missing a test for error 404 (client not found).

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: Yes Its more descriptive. Changed!

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandler.swift
ghuMustafa: Yap! I had created one but missed to throw the right one which is `invalidClientID`. Nice catch changed.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandlerTests.swift
ghuMustafa: Defiantly! I have added tests for these also.

Sources/Request Strategies/MLS/Actions/CountSelfMLSKeyPackagesActionHandlerTests.swift
ghuMustafa: Yes I somehow missed! I have added the test now.

johnxnguyen (approved):
👍🏻

PR #532, opened 2022-06-17
chore: create action handler test base

David-Henner:
----

- The **PR Title**
  - [ ] contains a reference JIRA issue number like `SQPIT-764`

- The **PR Description**
----

# What's new in this PR?

- Created a `ActionHandlerTestBase` class containing helpers methods to be used to test action handlers
- Updated following tests to use `ActionHandlerTestBase`:
  - `ClaimMLSKeyPackageActionHandlerTests`
  - `UploadSelfMLSKeyPackagesActionHandlerTests`
  - `SendMLSWelcomeActionHandlerTests`

### To do

- [x]  apply the same changes to the remaining action handler tests

### Possible improvements

1. Add a helper method for `test_itGeneratesARequest` (see [here](https://github.com/wireapp/wire-ios-request-strategy/blob/5c188fa891369c8f720fbfcf4b4358337b134ed1/Sources/Request%20Strategies/MLS/Actions/ClaimMLSKeyPackageActionHandlerTests.swift#L39))
2. Add a helper method to validate successful responses containing a payload (for tests like [this one](https://github.com/wireapp/wire-ios-request-strategy/blob/5c188fa891369c8f720fbfcf4b4358337b134ed1/Sources/Request%20Strategies/MLS/Actions/ClaimMLSKeyPackageActionHandlerTests.swift#L81))
3. Validate the action result in a more generic way

```swift
// Instead of having this
test_itHandlesResponse(status: 400) {
     guard case .failure(.invalidBody) = $0 else { return false } 
     return true
}

// We could do this
test_itHandlesFailure(status: 400, expectedError: .invalidBody)
```

In order to achieve this, we would need to make `EntityAction.Failure` conform to the `Equatable` protocol

----

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References


ghuMustafa (approved):

Tests/Helpers/ActionHandlerTestBase.swift
```diff
@@ -0,0 +1,113 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+@testable import WireRequestStrategy
+
+class ActionHandlerTestBase<Action: EntityAction, Handler: ActionHandler<Action>>: MessagingTestBase {
```
johnxnguyen, line 23: You may not need a generic parameter for the `Action`, as you should be able to get it from `ActionHandler.Action`.

Tests/Helpers/ActionHandlerTestBase.swift
```diff
@@ -0,0 +1,113 @@
...
+        var payload: [String: String]?
+        if let label = label {
+            payload = ["label": label]
+        }
+
+        return ZMTransportResponse(
+            payload: payload as ZMTransportData?,
+            httpStatus: status,
+            transportSessionError: nil,
+            apiVersion: APIVersion.v1.rawValue
```
johnxnguyen, line 110: Not a problem now, but hardcoding the api version may be a source of errors in the future if any of the actions are only available from a higher api version.

johnxnguyen (commented):
Nice work! Will the todo's happen in another PR?

Tests/Helpers/ActionHandlerTestBase.swift
David-Henner: Since `ActionHandler.Action` is also a generic, and I need to specify its type when declaring `Handler: ActionHandler<Action>`, I think I have no other choice than to declare `Action: EntityAction` first.

Tests/Helpers/ActionHandlerTestBase.swift
David-Henner: fixed

johnxnguyen (approved):

PR #533, opened 2022-06-20
fix: notifications not arriving SQCORE-1318

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-1318" title="SQCORE-1318" target="_blank">BugSQCORE-1318</a>  [iOS] Notifications are not arriving
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Since the NSE was released, notifications sometimes don't arrive or arrive late.

### Causes

After a while the NSE gets in a state where it no longer operates correctly and fails to process notifications. The result of this is that no notifications are shown to the user. Late notifications are likely the user notifications scheduled from the main app when it periodically fetches and processes events. After some time, it seems that the NSE recovers and can process notifications again, but eventually it will get stuck again.

There are possible several causes to this, but they likely all related to the stability of the NSE. One such issue that I've found is a retain cycle which causes a memory leak for a several instances in the NSE. As iOS creates new instances of the NSE, these old and leaked instances will still exist in memory, and eventually our memory footprint will exceed the memory budget for the extension. At this point the NSE will probably be terminated.

Further evidence supporting this idea is that once the NSE is in this state, restarting the device will resolve the issue, as we likely enter a clean slate when restarting.

### Solutions

This PR removes a single retain cycle, which together this one other retain cycle in another framework (see https://github.com/wireapp/wire-ios-notification-engine/pull/26), accounts for all the memory leaks in the NSE.

#### Test Coverage

- There are no unit tests for this, but I think we should consider finding some automated tools (linter?) to check that delegates are always `weak`.
- I did manual testing however and found the issue resolved: over a period of 3 days, I received hundreds of messages and there was a notification for every single message.

----

----

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References


David-Henner (approved):

ghuMustafa (approved):

johnxnguyen:
release patch

zenkins:
Hello, @johnxnguyen! Merging PR for command **release patch** for PR #**533**. Will create a **patch** release after merged.
Next version is 267.2.2.

PR #549, opened 2022-07-12
feat(mls): process welcome message FS-544

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-544" title="FS-544" target="_blank">TaskFS-544</a>  [iOS] Process incoming welcome
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

When being added to an MLS conversation, there are two kinds of update events being sent to the user:
- `conversation.create` / `conversation.member-join` 
- `conversation.mls-welcome`

These events do not arrive in any specific order, so two scenarios have to be handled.
(_To learn about this in depth, take a look at the ticket_)

In this PR, we are handling these two kinds of events by processing their payload and interacting with core crypto to either forward it the welcome message, or ask if the welcome message has already been processed. In both cases, we update the `isPendingWelcomeMessage` property of the conversation. 

When receiving a `conversation.create` event, we also update the `mlsGroupID` and `messageProtocol` properties of the conversation.

### Changes

- Added `messageProtocol` and `mlsGroupID` properties to `Payload.Conversation` and `Payload.UpdateConversationMemberJoin`
- Created a new `Payload.UpdateConversationMLSWelcome` for `conversation.mls-welcome` events
- Created `MLSEventProcessor` class, dedicated to processing update event data related to MLS
- Created `MLSEventProcessing` protocol, useful to mock the `MLSEventProcessor` during tests
- Updated `PayloadProcessing+Conversation` to process the update events related to MLS

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-transport/pull/255
- [ ] https://github.com/wireapp/wire-ios-data-model/pull/1386

### Testing

- Added tests for `MLSEventProcessor`
- Added tests to `PayloadProcessing+ConversationTests`
- Added a `MockMLSEventProcessor` class

**Note:** some tests still need to be updated

----

----

----
##### References


Sources/Payloads/Processing/Helpers/MLSEventProcessorTests.swift
```diff
@@ -0,0 +1,170 @@
...
+                conversation: self.conversation,
+                protocol: "mls",
+                groupID: self.groupIdString,
+                context: self.syncMOC
+            )
+
+            // Then
+            XCTAssertEqual(self.conversation.messageProtocol, .mls)
+        }
+    }
```
David-Henner, line 98: TODO: Move this test to PayloadProcessing tests

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
```diff
@@ -173,6 +174,8 @@ extension Payload.Conversation {
         updateMembers(for: conversation, context: context)
         updateConversationTimestamps(for: conversation, serverTimestamp: serverTimestamp)
         updateConversationStatus(for: conversation)
+        updateMessageProtocol(for: conversation)
+        updateMLSStatus(for: conversation, context: context)
```
David-Henner, line 178: My assumption is that only group conversation are MLS conversations

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
```diff
@@ -173,6 +174,8 @@ extension Payload.Conversation {
         updateMembers(for: conversation, context: context)
         updateConversationTimestamps(for: conversation, serverTimestamp: serverTimestamp)
         updateConversationStatus(for: conversation)
+        updateMessageProtocol(for: conversation)
```
David-Henner, line 177: We may need to do this too for the `conversation.member-join` event. It's unclear to me if we should always receive a `conversation.create` event when a new conversation has been created.

johnxnguyen: As I understand this `Conversation` payload is used for many different things. Do you know if this payload is used for both event types?

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
johnxnguyen: For now I think that assumption is correct, but it won't hold forever.

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
```diff
@@ -496,3 +531,20 @@ extension Payload.ConversationEvent where T == Payload.UpdateConversationConnect
     }
 
 }
+
+extension Payload.ConversationEvent where T == Payload.UpdateConversationMLSWelcome {
+
+    func process(in context: NSManagedObjectContext, originalEvent: ZMUpdateEvent) {
+        guard
+            let identifier = id ?? qualifiedID?.uuid,
+            let domain = qualifiedID?.domain ?? APIVersion.domain
```
johnxnguyen, line 540: It would be good to check that `qualifiedID?.domain` is not empty. We should have a helper already, like `qualifiedID?.domain.nilIfEmpty`

Sources/Payloads/Processing/Helpers/MLSEventProcessor.swift
```diff
@@ -0,0 +1,79 @@
...
+    func process(welcomeMessage: String, for conversation: ZMConversation?, in context: NSManagedObjectContext) {
+        let groupID = context.mlsController?.processWelcomeMessage(welcomeMessage: welcomeMessage)
+
+        guard groupID != nil else {
+            return Logging.eventProcessing.error("Couldn't process welcome message")
+        }
+
+        conversation?.isPendingWelcomeMessage = false
+    }
+}
```
johnxnguyen, line 59: So we call this also when the conversation might not yet exist? And we don't need to use the group ID because we'll get it from the conversation create event?

David-Henner: > So we call this also when the conversation might not yet exist?

Yes, since it's possible that we receive a `conversation.mls-welcome` event before receiving the `conversation.create` or `conversation.member-join` event. At which point the conversation would not exist (at least locally) yet. But core crypto still needs to process the welcome message. 
Then once we receive the event about the conversation being created, we can ask core crypto if it already processed the welcome message for the conversation and update the `isPendingWelcomeMessage` field accordingly

> And we don't need to use the group ID because we'll get it from the conversation create event?

Correct, we don't need to use the group ID here. We also get it from the conversation create event, which we then use to ask core crypto if it processed the welcome message for that conversation. 
The only use for groupID in this method above, would be to let us fetch the conversation, but we can already do that from the conversation remote identifier that we received from the event.

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
David-Henner: For `conversation.member-join`, the `Payload.UpdateEventConversationMemberJoin` is used. Each of these payload types are processing the event differently (rightly so). But only when processing the `conversation.create` event do we set the all the conversation properties. Now say that the `conversation.member-join` is received, but we don't have a local conversation, we'd need to create a new one, and some of the conversation properties wouldn't be set. 

So, we may want to make sure that, in this case, the message protocol is also set for the conversation

Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
David-Henner: [1021f88](https://github.com/wireapp/wire-ios-request-strategy/pull/549/commits/1021f88a7fca3ceae633d205480339e2bf0ef327)

johnxnguyen (approved):

ghuMustafa (approved):

PR #554, opened 2022-07-28
feat(mls): decrypt messages FS-650

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-650" title="FS-650" target="_blank">TaskFS-650</a>  [iOS] Receive an MLS message
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR introduces support for decrypting incoming `conversation.mls-message-add` events. 

The main challenge here was to find the right place in the existing message processing flow to add the decryption of the encrypted mls message data. 

The `EventDecoder` is where OTR messages are already decrypted with cryptobox. There we extract the encrypted message data from the `ZMUpdateEvent` instances payload, decrypt it, and creates a new `ZMUpdateEvent` instance copying the original update event, and inserting the decrypted message data into the payload. 

For MLS messages, we essentially will do the same process, using core crypto instead of cryptobox to do the decryption. 

### Changes

- Extension `ZMUpdateEvent+Decryption` will take care of inserted decrypted data into the payload
- Extension `ZMUpdateEvent+Payload` is a helper to convert payload data into a `Codable` struct
- Added `UpdateConversationMLSMessageAdd` to represent the payload of `conversation.mls-message-add` events
- Added support of `conversationMLSMessageAdd` event types to `ClientMessageRequestStrategy`
- Added logic for decrypting the event in `EventDecoder`

### Dependencies

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/1392

### Testing

- Added tests to `EventDecoderTest` to cover the logic of update event decryption
- Minor refactoring for `MockMLSController`

----

----

----
##### References


Sources/Request Strategies/Client Message/ClientMessageRequestStrategy.swift
```diff
@@ -125,31 +125,20 @@ extension ClientMessageRequestStrategy: ZMEventConsumer {
...
-            case .conversationClientMessageAdd, .conversationOtrMessageAdd, .conversationOtrAssetAdd:
-                if let nonce = $0.messageNonce {
-                    return UpdateEventWithNonce(event: $0, nonce: nonce)
-                }
-                return nil
-            default:
-                return nil
-            }
-        }
-    }
```
David-Henner, line 148: this was marked as `private` and was never called in the scope

Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -140,6 +143,42 @@ extension EventDecoder {
...
+        }
+
+        guard let payload = updateEvent.eventPayload(type: Payload.UpdateConversationMLSMessageAdd.self) else {
+            Logging.eventProcessing.warn("invalid update event payload")
+            return nil
+        }
+
+        // NOTE: May need to change this so that we get the group ID from the payload in case there's a handshake message
+        // that needs to be processed before the conversation has been created
+        guard let conversation = ZMConversation.fetch(with: payload.id, domain: payload.qualifiedID?.domain, in: context) else {
```
David-Henner, line 159: unsure if we should fall back on the `APIVersion.domain` here

Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -140,6 +143,43 @@ extension EventDecoder {
...
+            return nil
+        }
+
+        guard let payload = updateEvent.eventPayload(type: Payload.UpdateConversationMLSMessageAdd.self) else {
+            Logging.eventProcessing.warn("invalid update event payload")
+            return nil
+        }
+
+        // NOTE: May need to change this so that we get the group ID from the payload in case there's a handshake message
+        // that needs to be processed before the conversation has been created
```
David-Henner, line 158: Note: this is still an open question

Sources/Helpers/ZMUpdateEvent+Decryption.swift
```diff
@@ -0,0 +1,88 @@
...
+extension ZMUpdateEvent {
+
+    // MARK: MLS Message Events
+
+    /// Creates a new instance of `ZMUpdateEvent` replacing the encrypted event data with the decrypted data passed as parameter
+    /// - Parameter decryptedData: data representing the decrypted value of the update event data. Must have been decrypted with core crypto
+    /// - Returns: a version of `self` with a payload containing the decrypted data
+    func decryptedMLSEvent(decryptedData: Data) -> ZMUpdateEvent? {
+        guard type == .conversationMLSMessageAdd else {
+            fatal("decrypting wrong type of event")
```
johnxnguyen, line 30: Do you think we should crash or could we also return nil?

Sources/Synchronization/Decoding/EventDecoder.swift
johnxnguyen: I think if there is no domain, then it will assume the local domain anyway.

johnxnguyen (approved):

Sources/Helpers/ZMUpdateEvent+Decryption.swift
David-Henner: `fatal` is maybe to severe so we could use an `assert`. The purpose of using this instead of `return nil` is to tell developers that this method is designed to be used only for `.conversationMLSMessageAdd` events

ghuMustafa (approved):

PR #564, opened 2022-09-07
fix: isPendingWelcomeMessage is true for the group creator FS-545

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-545" title="FS-545" target="_blank">TaskFS-545</a>  [iOS] Send message in an MLS group
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When user A creates an mls group with user B, A can not receive messages sent successfully from B.

### Causes

The logs indicate that for user A, they receive the message but can not decrypt it because conversation is pending a welcome message. Since A created the group, no welcome message is needed, so this state is incorrect.

After creating the conversation on the backend, we create/update the conversation locally in the backend, then create the mls group in Core Crypto. Part of updating the conversation locally means to set the mls group ID and set the flag `isPendingWelcomeMessage` to `true` if the mls group doesn't yet exist in Core Crypto. Since this happens before we create the group in Core Crypto, we set the value to `true`.

### Solutions

Reset `isPendingWelcomeMessage` to `false` before creating the group in Core Crypto. This only happens when the conversation has been created by the self client.

#### Test Coverage (Optional)

- Added an assertion that `isPendingWelcomeMessage` is `false` after creating an mls group.

----

----

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References


Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -378,6 +382,9 @@ extension ConversationRequestStrategy: ZMUpstreamTranscoder {
         newConversation.needsToBeUpdatedFromBackend = deletedDuplicate
 
         if newConversation.messageProtocol == .mls {
+            Logging.mls.info("resetting `isPendingWelcomeMessage` to `false` b/c self client is creator")
+            newConversation.isPendingWelcomeMessage = false
```
David-Henner, line 386: note that this will need to be changed for `mlsStatus` since it replaces `isPendingWelcomeMessage`

David-Henner (approved):

